PR #1, opened 2021-11-08
feat: implement skeleton of osd-network-verifier

boranx:
Implements: https://issues.redhat.com/browse/OSD-8881

```
./osd-network-verifier byovpc
interface executed: AWS
success
```


cmd/byovpc/cmd.go
```diff
@@ -0,0 +1,25 @@
...
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+func NewCmdByovpc(streams genericclioptions.IOStreams) *cobra.Command {
+	byovpcCmd := &cobra.Command{
+		Use: "byovpc",
+		Run: func(cmd *cobra.Command, args []string) {
+			// Do Stuff Here
+			// AWS & GCP
+			// cloudclient.GetClientFor(configv1.AWSPlatformType)
```
boranx, line 16: platform type will be decided and put here

bng0y (approved):

cmd/byovpc/cmd.go
```diff
@@ -0,0 +1,42 @@
+package byovpc
+
+import (
+	"context"
+	"os"
+
+	configv1 "github.com/openshift/api/config/v1"
+	cloudclient "github.com/openshift/osd-network-verifier/pkg"
```
tiwillia, line 8: New to the team so I may not be up-to-date on code conventions/organization, but it seems to me like `pkg/cloudclient/` should contain the files for cloudclient (pkg/add_aws.go, pkg/add_gcp.go, etc) and this should be importing `github.com/openshift/osd-network-verifier/pkg/cloudclient` un-named.

Generally there are no top-level files in pkg/ for most golang projects AFAIK.

boranx: yeah, that's a valid concern. However, I believe due to https://github.com/openshift/osd-network-verifier having no pkg/
It's giving me compilation error if I wanna say something like:
```shell
cloudclient "github.com/openshift/osd-network-verifier/pkg/cloudclient"
```
```
could not import github.com/openshift/osd-network-verifier/pkg/cloudclient (no required module provides package "github.com/openshift/osd-network-verifier/pkg/cloudclient")compilerBrokenImport
```
Is there a way to workaround this?

boranx: Hey @tiwillia, yeah, you're right... Seems that my mistake was not adding `cloudclient.go` to the bundle :detective: 
I've just applied a fix to apply your changes, thanks! great suggestions :)

fahlmant (approved):
/lgtm

PR #12, opened 2021-12-06
Simplifying init for cloudclient

bng0y:
cloudclient is now built with `cloudclient.NewClient(credentials interface{}, region string)`
which will build:
- a aws client if the type of `crendentials` is `credentials.StaticCredentialsProvider` from aws-sdk-go-v2
- a gcp client if the type of `credentials` is `*google.Credentials`


robotmaxtron (approved):
/lgtm

PR #23, opened 2021-12-09
OSD-9228 unit tests

ritmun:
Do not merge yet. Draft PR for unit tests.


fahlmant (commented):
The vendor part can be removed now that Boran has removed the need for it.

PR #26, opened 2021-12-13
Use ocm-sdk-go logger interface and stdout implementation 

tiwillia:
This change introduces the ocm-sdk-go logging interface to the project.
The primary consumer of this tool will be uhc-clusters-service, which
appears to use this interface for all of its logging. Using the same
interface will ensure the logger already used by uhc-clusters-service
can simply be passed into the cloudclient library for easy usage.

There are several implementations in the ocm-sdk-go library of the
interface, one of which we use here as an extremely basic implementation
of output to stdout/stderr. If we want to move to zap or glog or any
other logging library, there are good examples to adapt to these
implementations using the interface.

A new command-line boolean `--debug=true` is added.

Debug output will now include the base64-encoded generated userdata
script, as well as the complete ec2 userdata output for review in
development.


tiwillia:
An additional change that was made as part of these additions was adding a `context.Context` parameter as the first parameter to most functions. This is standard practice even though the context is not explicitly used by anything other than the logger.

fahlmant (approved):
/approve

PR #27, opened 2021-12-14
Wait for verification that userdata script has completed and move to nitro instance type

tiwillia:
This includes several changes due to an issue that was revealed while adding the initial change to properly check for userdata completion:

1. Userdata script completion checking is added/changed to use regexp
2. Moving the default instance type to use a "nitro" t3.micro instance that supports continuous console log output
3. Providing the `latest=true` request parameter to the GetConsoleOutput EC2 API call to ensure the console log is updated before being provided in the response
4. Adding a default region `us-east-2` instead of assuming the AWS_DEFAULT_REGION variable is defined. We may want to consider add this as a command-line option as well.

This resolves a major issue where console output may not include the full userdata execution, depending on how long the userdata script takes to execute. After these changes, the egress check will execute reliably.

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -36,7 +36,14 @@ func NewCmdValidateEgress() *cobra.Command {
 			ctx := context.TODO()
 
 			creds := credentials.NewStaticCredentialsProvider(os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), os.Getenv("AWS_SESSION_TOKEN"))
-			region := os.Getenv("AWS_DEFAULT_REGION")
+			// TODO this should probably be a command-line option that overwrites the env var
+			regionEnvVarStr := "AWS_DEFAULT_REGION"
+			regionDefault := "us-east-2"
+			region := os.Getenv(regionEnvVarStr)
+			if len(region) < 1 {
+				logger.Warn(ctx, "No region defined in %s env, defaulting to %s", regionEnvVarStr, regionDefault)
+				region = regionDefault
+			}
 
 			cli, err := cloudclient.NewClient(ctx, logger, creds, region, cloudTags)
 			if err != nil {
--- a/pkg/cloudclient/aws/private.go
+++ b/pkg/cloudclient/aws/private.go
@@ -46,6 +46,7 @@ var (
 		"af-south-1":     "",
 		"me-south-1":     "",
 	}
+	userdataEndVerifier string = "USERDATA END"
 )
 
 func newClient(ctx context.Context, logger ocmlog.Logger, accessID, accessSecret, sessiontoken, region string, tags map[string]string) (*Client, error) {
@@ -207,7 +208,7 @@ func generateUserData() (string, error) {
 	// TODO find a location for future docker images
 	data.WriteString("sudo docker pull docker.io/tiwillia/network-validator-test:v0.1\n")
 	data.WriteString("sudo docker run docker.io/tiwillia/network-validator-test:v0.1\n")
-	data.WriteString("echo 'USERDATA END' \n")
+	data.WriteString(fmt.Sprintf(`echo "%s"`, userdataEndVerifier) + "\n")
 
 	userData := base64.StdEncoding.EncodeToString([]byte(data.String()))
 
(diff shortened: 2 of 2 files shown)


pkg/cloudclient/aws/private.go
```diff
@@ -220,14 +221,30 @@ func (c Client) findUnreachableEndpoints(ctx context.Context, instanceID string)
...
+			if len(scriptOutput) < 1 {
+				c.logger.Debug(ctx, "EC2 console output not yet populated with data, continuing to wait...")
+				return false, nil
+			}
+
+			// Check for the specific string we output in the gerated userdata file at the end to verify the userdata script has run
+			// It is possible we get EC2 console output, but the userdata script has not yet completed.
+			reVerify := regexp.MustCompile(userdataEndVerifier)
+			verifyMatch := reVerify.FindString(string(scriptOutput))
+			if len(verifyMatch) < 1 {
```
tiwillia, line 242: As things stand, if we hit this we should probably just error out, as it seems the log won't ever be updated if the first update does not include the userdata output.

pkg/cloudclient/aws/private.go
```diff
@@ -217,26 +220,45 @@ func generateUserData() (string, error) {
...
 
-			// If debug logging is enabled, output the full console log
-			c.logger.Debug(ctx, "Full EC2 console output:\n---\n%s\n---", scriptOutput)
+			// In the early stages, an ec2 instance may be running but the console is not populated with any data, retry if that is the case
+			if len(scriptOutput) < 1 {
+				c.logger.Debug(ctx, "EC2 console output not yet populated with data, continuing to wait...")
+				return false, nil
+			}
 
-			if !strings.Contains(string(scriptOutput), "USERDATA END") {
```
tiwillia, line 233: Replaced this check @bng0y added with a regexp and ensured we use a single variable to define the string we are looking for, used in both places we rely on it so if it ever changes, both are updated.

cmd/egress/cmd.go
```diff
@@ -36,7 +36,14 @@ func NewCmdValidateEgress() *cobra.Command {
 			ctx := context.TODO()
 
 			creds := credentials.NewStaticCredentialsProvider(os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), os.Getenv("AWS_SESSION_TOKEN"))
-			region := os.Getenv("AWS_DEFAULT_REGION")
+			// TODO this should probably be a command-line option that overwrites the env var
```
fahlmant, line 39: Mind creating an OSD card for this? +1 to making these things configurable

pkg/cloudclient/aws/private.go
```diff
@@ -20,7 +20,8 @@ import (
 )
 
 var (
-	instanceType  string = "t2.micro"
+	// NOTE a "nitro" EC2 instance type is required to be used
+	instanceType  string = "t3.micro"
```
fahlmant, line 24: Just a though; I wonder if there's a way to see if an instance type is "nitro" or not from the AWS API. Then we could make the instance type configurable as long as its "nitro"

fahlmant (approved):
/lgtm

pkg/cloudclient/aws/private.go
tiwillia: There looks to be a `describe instance type` API that we could utilize to check whether the provided instance type is indeed nitro by reviewing the instance type's `hypervisor` attribute: https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-types.html

Is this worth putting together an OSD card for?

cmd/egress/cmd.go
tiwillia: Sure thing: https://issues.redhat.com/browse/OSD-9344

pkg/cloudclient/aws/private.go
tiwillia: https://issues.redhat.com/browse/OSD-9345

pkg/cloudclient/aws/private.go
```diff
@@ -206,8 +208,9 @@ func generateUserData() (string, error) {
 	data.WriteString("sudo service docker start\n")
 	// TODO find a location for future docker images
 	data.WriteString("sudo docker pull docker.io/tiwillia/network-validator-test:v0.1\n")
-	data.WriteString("sudo docker run docker.io/tiwillia/network-validator-test:v0.1\n")
-	data.WriteString("echo 'USERDATA END' \n")
+	// Use `|| true` to ignore failure exit codes, we want the script to continue either way
+	data.WriteString("sudo docker run docker.io/tiwillia/network-validator-test:v0.1 || true\n")
```
bng0y, line 212: this is cool

PR #29, opened 2021-12-15
[OSD-9042] Small Readme updates, GCP unit tests

robotmaxtron:
Some small linting changes added to README, GCP unit tests added.

### DIFF
--- a/pkg/cloudclient/gcp/gcp_test.go
+++ b/pkg/cloudclient/gcp/gcp_test.go
@@ -1,0 +1,58 @@
+package gcp
+
+import (
+	"context"
+	"testing"
+
+	gomock "github.com/golang/mock/gomock"
+	mock_cloudclient "github.com/openshift/osd-network-verifier/pkg/cloudclient/mock_cloudclient"
+
+	ocmlog "github.com/openshift-online/ocm-sdk-go/logging"
+	"golang.org/x/oauth2/google"
+)
+
+func TestByoVPCValidator(t *testing.T) {
+	ctx := context.TODO()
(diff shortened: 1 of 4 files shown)


pkg/cloudclient/gcp/gcp_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+
+	gomock "github.com/golang/mock/gomock"
+	mock_cloudclient "github.com/openshift/osd-network-verifier/pkg/cloudclient/mock_cloudclient"
+
+	ocmlog "github.com/openshift-online/ocm-sdk-go/logging"
+	"golang.org/x/oauth2/google"
+)
+
+func TestByoVPCValidator(t *testing.T) {
```
boranx, line 14: This is actually not testing ByoVPCValidator's logic itself. Because we're mocking what we expect from the output,  err `err := mock.ByoVPCValidator(ctx)` the err here would always be nil.

This pattern actually can be used when calling **ByoVPCValidator** externally from cmd functions, like https://github.com/openshift/osd-network-verifier/blob/main/cmd/byovpc/cmd.go#L36
If this was the case, Instead of calling ByoVPCValidator, we could call `mock.ByoVPCValidator(ctx)` which would return nil so that we could check the functionality of `NewCmdByovpc() *cobra.Command`

If we want to test out internal functionality, we could use something like that:
```go
func TestByoVPCValidator(t *testing.T) {
	ctx := context.TODO()
	cli := Client{logger: &ocmlog.GlogLogger{}}
	if err := cli.ByoVPCValidator(ctx); err != nil {
		t.Errorf("ByoVPCValidator() should return nil")
	}
}
```

PR #30, opened 2021-12-16
cross-platform userdata

bng0y:
- Allow the region to be passed by command line `--region` [OSD-9344](https://issues.redhat.com/browse/OSD-9344)
- Uses [cloud-config](https://cloudinit.readthedocs.io/en/latest/topics/examples.html) syntax for userdata script
- Validating reachability to region specific resource. e.g ec.<region>.amazonaws.com [OSD-9222](https://issues.redhat.com/browse/OSD-9222)


tiwillia:
One comment, the rest of this LGTM

cmd/egress/cmd.go
```diff
@@ -15,11 +15,23 @@ var defaultTags = map[string]string{
 	"osd-network-verifier": "owned",
 }
 
+func getDefaultRegion() string {
```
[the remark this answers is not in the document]
bng0y: interestingly, it shows up in the `./osd-network-verifier --help`
```
--region string               Region to validate (default "us-east-2")
```
or 
`default "AWS_DEFAULT_REGION"` when it is set

fahlmant:
@bng0y Could you please add a description?

cmd/egress/cmd.go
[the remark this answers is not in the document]
bng0y: I returned the `logger.Warn(ctx, "Using region: %s", region)` and updated the `--help` with more explanation

config/userdata.yaml
```diff
@@ -0,0 +1,13 @@
+#cloud-config
+repo_update: true
+
+packages:
+  - docker 
+
+runcmd:
+  - [ sh, -c, echo "USERDATA BEGIN"]
+  - sudo service docker start
```
tiwillia, line 9: Just wanted to note that this removes `yum update` and `yum install docker`.

bng0y: the `packages: docker` covers the latter.

bng0y: we could consider also adding `package_update: true`

tiwillia: James Harrington confirmed in slack that we did not need to worry about doing `yum update`, so fine to leave `package_update: true` out fro now I think.

tiwillia:
LGTM. Will conflict with https://github.com/openshift/osd-network-verifier/pull/31 , lets merge this one first imo.

fahlmant (approved):
/approve

PR #33, opened 2021-12-21
feat: apply testing suite

boranx:
This PR does a few things:

- Fix Broken gcp test: `TestNewClient()`
- Fix gcp/TestValidateEgress() which was not covering the actual function but was testing against mock. (This'll increase code coverage)
- Initialize mocking structure for AWS SDK(ec2 client), Write test against createEC2Instance which is internally calling aws ec2 SDK that needs to be mocked 
- Apply the first test for aws package

**P.s:** The content in this PR will be demo-ed in a recorded session


fahlmant (approved):
/lgtm

PR #37, opened 2022-01-14
[OSD-9584] Allow timeout to be specified from osd-network-verifier CLI

tiwillia:
A timeout for each individual egress request is already supported by the
network-validator binary. Prior to this change, using the binary through
the osd-network-validator tool resulted in the timeout being hard-coded
to 1 second. This change allows the timeout to be specified via the
osd-network-verifier CLI or via ValidateEgress cloudclient function call

Note that this will require pushing a new docker image version of the
network-validator image, simply to change the ENTRYPOINT in the
dockerfile to allow arguments to be passed through the run of the image.


build/Dockerfile
```diff
@@ -26,4 +26,4 @@ WORKDIR /app
 COPY --from=builder /app/bin/network-validator /usr/bin/network-validator
 COPY --from=builder /app/config/ /app/config/
 
-ENTRYPOINT network-validator --config=/app/config/config.yaml --timeout=1s
+ENTRYPOINT ["network-validator", "--config=/app/config/config.yaml"]
```
tiwillia, line 29: This change was necessary to allow additional command-line options for the network-validator to be passed when running the docker image. This will require a new version of the docker image to be pushed.

tiwillia: @boranx I think pushing a new version of the docker image is either something you've automated or handle manually? Could you PTAL?

boranx: @tiwillia that's a nice change!, I've tested it on my local as well
We can locally produce the image via `make containerbuild` but it's not onboarded to _quay.io/app-sre_ yet
Afaik the current image is still hard-coded at https://github.com/openshift/osd-network-verifier/blob/main/pkg/cloudclient/aws/private.go#L50

pkg/cloudclient/aws/private.go
```diff
@@ -47,7 +47,7 @@ var (
 		"me-south-1":     "",
 	}
 	// TODO find a location for future docker images
-	networkValidatorImage string = "quay.io/bngoy/osd-network-verifier:1"
+	networkValidatorImage string = "quay.io/twilliams725/osd-network-verifier:2"
```
tiwillia, line 50: Pushed a new image based on these changes, this updates the project to use the new image until we have better automation around image CD in place.

fahlmant (approved):
/lgtm

PR #45, opened 2022-02-02
Implement local PollImmediate to remove dependency on k8s.io/apimachinery

fahlmant:
Adding a local implementation of the PollImmediate logic to remove unnecessary dependency on `k8s.io/apimachinery`


pkg/helpers/helpers.go
```diff
@@ -0,0 +1,26 @@
...
+		cond, err := condition()
+		if cond {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		time.Sleep(interval)
+		totalTime += interval
+		fmt.Printf("totalTime: %d, timeout: %d\n", totalTime, timeout)
```
rafael-azevedo, line 22: Did you mean to leave this in or was it debug ?

pkg/helpers/helpers.go
```diff
@@ -0,0 +1,25 @@
...
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		time.Sleep(interval)
+		totalTime += interval
+	}
+
+	return fmt.Errorf("timed out")
```
tiwillia, line 24: A typed error would enable the caller to reliably determine whether the failure here was due to a timeout or an error returned from the condition func. See https://github.com/kubernetes/apimachinery/blob/276a8a7530a30f42c82b1f0fa2261a905130bf97/pkg/util/wait/wait.go#L672 for example.

tiwillia:
LGTM, just one suggested enhancement that can certainly be pushed off.

PR #48, opened 2022-02-10
embed userdata.yaml file

bng0y:
Embeds userdata.yaml file instead of looking for it locally


boranx (approved):
This looks neat to me, tested #47 including these changes, the validateegress test is passing
/lgtm /hold
cc: @fahlmant any last words before we go?

PR #49, opened 2022-02-17
Cleaned up summary

fahlmant:
Small change that only prints errors if they occur, otherwise prints that all tests pass


boranx:
/lgtm
Nice tweak :)

PR #50, opened 2022-02-17
Move to net/http connection test

fahlmant:
The egress tests currently check if a connection can be established with a specified endpoint. However, with layer 7 packed inspection, the connection is allowed to establish but packets are still filtered. This causes the verifier to show egress urls as reachable when no data is able to pass in cases where layer 7 filtering is used. This PR changes the test to check the response instead of just the connection to more accurately reflect the reachability of the list of urls.

https://issues.redhat.com/browse/OSD-10903

### DIFF
--- a/build/bin/network-validator.go
+++ b/build/bin/network-validator.go
@@ -13,8 +13,6 @@ import (
 	"sync"
 	"time"
 
-	"crypto/tls"
-
 	"golang.org/x/crypto/ssh"
 
 	"gopkg.in/yaml.v2"
@@ -105,9 +103,6 @@ func ValidateReachability(host string, port int) error {
 	endpoint := fmt.Sprintf("%s:%d", host, port)
 	httpClient := http.Client{
 		Timeout: *timeout,
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-		},
 	}
 
 	fmt.Printf("Validating %s\n", endpoint)
--- a/build/bin/network-validator.go
+++ b/build/bin/network-validator.go
@@ -7,11 +7,15 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"net"
+	"net/http"
 	"os"
 	"sync"
 	"time"
 
-	"net/http"
+	"crypto/tls"
+
+	"golang.org/x/crypto/ssh"
 
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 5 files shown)


build/bin/network-validator.go
```diff
@@ -86,11 +87,29 @@ func TestEndpoints(config reachabilityConfig) {
 }
 
 func ValidateReachability(host string, port int) error {
-	endpoint := fmt.Sprintf("%s:%d", host, port)
+	var endpoint string
+
+	httpClient := http.Client{
+		Timeout: *timeout * time.Second,
```
bng0y, line 93: I believe you can skip the `* time.Second` here given the type of `timeout`

fahlmant: +1

build/bin/network-validator.go
```diff
@@ -86,11 +87,29 @@ func TestEndpoints(config reachabilityConfig) {
...
+		endpoint = fmt.Sprintf("%s://%s", "http", host)
+	}
+
 	fmt.Printf("Validating %s\n", endpoint)
-	_, err := net.DialTimeout("tcp", endpoint, *timeout)
+
+	_, err := httpClient.Get(endpoint)
 	if err != nil {
-		return fmt.Errorf("Unable to reach %s within specified timeout: %s", endpoint, err)
+		return fmt.Errorf("unable to reach %s within specified timeout: %s", endpoint, err)
```
boranx, line 111: would this require a change also in the comparison mechanism that we're performing while validation?
https://github.com/openshift/osd-network-verifier/blob/main/pkg/cloudclient/aws/private.go#L261

abyrne55:
Code itself looks good to me, but due to the capitalization change (`Unable` vs. `unable`), won't we have to immediately follow up this PR with another PR that updates [the quay.io tag that we're pointing to](https://github.com/openshift/osd-network-verifier/blob/b71e3eac178340e476a2d3a6894813115d15b3e0/pkg/cloudclient/aws/private.go#L61)?

Asking because I noticed while testing this PR that reachability errors were being missed because the docker image with the "old" network-verifier (that outputs `unable...`) wasn't producing the output expected by the "new" osd-network-verifier binary (which expects `Unable...`).

We could also avoid all of this by making [this regex check](https://github.com/openshift/osd-network-verifier/blob/f6769f98117dead726e86612ee91eef45f2b0271/pkg/cloudclient/aws/private.go#L261) case-insensitive

boranx:
That's a very good point @abyrne55 Why do we want to change the letter to lower case? Is there a specific reason for it?
I'd prefer not to touch it until it gets case-sensitive form

fahlmant:
Go linting prefers Error types to start with a lowercase letter. I can change it back

build/bin/network-validator.go
```diff
@@ -86,11 +91,35 @@ func TestEndpoints(config reachabilityConfig) {
 }
 
 func ValidateReachability(host string, port int) error {
+	var err error
 	endpoint := fmt.Sprintf("%s:%d", host, port)
+	httpClient := http.Client{
+		Timeout: *timeout,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
michaelryanmcneill, line 99: We likely do not want to do this. If we disable TLS verification for all endpoints, it could (and likely will) still result in the cluster failing to provision. The only endpoints that need TLS inspection disabled are the four domains here https://docs.openshift.com/rosa/networking/configuring-cluster-wide-proxy.html#cluster-wide-proxy-network-prereqs_cluster-wide-proxy-configuration. The remainder should absolutely verify that they can connect using trusted certificates or the trust bundle provided by the end user.

michaelryanmcneill (changes requested):

build/bin/network-validator.go
fahlmant: :+1: fixing

michaelryanmcneill:
Love this most recent change @fahlmant! Will definitely make things easier as we add future endpoints.

ritmun (approved):

fahlmant:
/hold cancel

Thanks all!

PR #52, opened 2022-03-24
Error when the timeout condition is hit

fahlmant:
If the timeout condition for PollImmediate is hit, now the error will be reported properly.


sam-nguyen7 (approved):
/lgtm

pkg/cloudclient/aws/private.go
```diff
@@ -378,7 +380,11 @@ func (c *Client) validateEgress(ctx context.Context, vpcSubnetID, cloudImageID s
 	}
 
 	c.logger.Info(ctx, "Gathering and parsing console log output...")
-	c.findUnreachableEndpoints(ctx, instanceID)
+	err = c.findUnreachableEndpoints(ctx, instanceID)
+	if err != nil {
+		fmt.Println(err)
```
boranx, line 385: c.logger.Error(ctx, err.Error())

boranx: tbh, we don't even need to print out the err here as it'd be printed in the context of `output` anyway

PR #58, opened 2022-04-20
Allow users to specify instance root volume encryption key

abyrne55:
Added a `--kms-key-id <string>` CLI flag that allows the user to specify which KMS key should be used when encrypting the root volume of ONV-created instances. This option has also been added as a parameter (`kmsKeyID`) to the `ValidateEgress()` API function. Not specifying this key simply defaults back to the cloud account default KMS key.

Implementation-wise, adding this option as yet-another parameter to `createEC2Instance()` would have made that function's signature unreadably long, so I instead refactored the function such that it now takes in a param struct of type createEC2InstanceInput.

This addresses OSD-11084, where verifier users requested the ability to specify which KMS key is used for encryption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ cli, err := cloudclient.NewClient(creds, region)
 // ... error checking
 
 // call the validation function and check if it was successful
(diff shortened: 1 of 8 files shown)


fahlmant (approved):
/approve

PR #59, opened 2022-04-25
osd-10310: Creating AWS infrastructre templates for osd-network-verifier testing 

lnguyen1401:
PR content:
a go script to create a VPC and firewall for OSD-network-verifier testing.
How to test:
run aws sts assume-role --role-arn arn:aws:iam::"$your-aws-account-number":role/OrganizationAccountAccessRole --role-session-name "$your-name" --profile osd-staging-2
run go builld awsconfig.go
AWS_ACCESS_KEY_ID= AWS_SECRET_ACCESS_KEY= AWS_SESSION_TOKEN= ./firewallConfig
open your AWS console and see that VPC and Firewall have been created.


pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			default:
+				fmt.Println(aerr.Error())
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+	time.Sleep(wait)
```
abyrne55, line 269: Could you explain why there's a 2-minute wait here? If we're just waiting on AWS to bring up the NAT Gateway, maybe indicate that in the console output? When I ran into this delay initially, I assumed the script crashed.

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			default:
+				fmt.Println(aerr.Error())
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
```
abyrne55, line 99: Perhaps this is a Golang convention that I'm not familiar with, but I'm having a lot of trouble reading these heavily-nested error-handling blocks. Is there simpler logic that could be used here?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	fmt.Println("Successfully created a Firewall. Having to wait a long time for the endpoint to be ready!")
```
abyrne55, line 367: See above. I'd suggest `Successfully created a Firewall. Waiting 6 minutes for endpoint to become ready...`

abyrne55 (commented):
Built using `go build pkg/awsconfig/testing/firewallConfig.go` and successfully ran the resulting binary, verifying that the VPC, IG, NAT gateway, firewall, and firewall policy were properly created. Great work!

I left a few comments about readability and output. I'm assuming more thorough are coming soon since this is a WiP.

pkg/awsconfig/testing/firewallConfig.go
fahlmant: Any Sleeps we should refactor to retries with backoff

pkg/awsconfig/testing/firewallConfig.go
fahlmant: The else can be dropped in 96. And unless a specific AWS error code is being handled in the switch case, that switch can also be dropped.

pkg/awsconfig/testing/firewallConfig.go
lnguyen1401: Yeah that's to wait for the NAT Gateway to be ready. will add a log saying this.

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/aws/aws-sdk-go/service/networkfirewall"
+	"github.com/aws/aws-sdk-go/service/networkfirewall/networkfirewalliface"
+)
+
+type Client struct {
+	ec2Client ec2iface.EC2API
+	firewallClient networkfirewalliface.NetworkFirewallAPI
+}
+var wait = 120 * time.Second
```
fahlmant, line 22: Please add a blank new line between the Client and this var

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+var wait2 = 360 * time.Second
+func main() {
```
fahlmant, line 24: Please add a new line before this

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+
+	creds := credentials.NewStaticCredentials(os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), os.Getenv("AWS_SESSION_TOKEN"))
```
fahlmant, line 26: What if someone wants to pass in an AWS profile instead?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	region := "us-east-1"
```
fahlmant, line 27: What if someone wants to use a different region?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	//Create Firewall
+	Firewall := awsClient.CreateFirewall(FirewallSubnet, Vpc)
+	//Describe Firewall
+	DescribeFirewall :=awsClient.DescribeFirewall(Firewall)
+	//Create route 0.0.0.0/0 in PublicRT for FirewallEndpoint
+	awsClient.CreateRouteToFirewall("0.0.0.0/0", *DescribeFirewall.FirewallStatus.SyncStates[*FirewallSubnet.Subnet.AvailabilityZone].Attachment.EndpointId, PublicRT)
+	fmt.Println("Successfully route 0.0.0.0/0 to the Firewall Endpoint in PublicRT ")
+	//Create route 10.0.0.0/24 in IgRt to FirewallEndpoint
+	awsClient.CreateRouteToFirewall("10.0.0.0/24", *DescribeFirewall.FirewallStatus.SyncStates[*FirewallSubnet.Subnet.AvailabilityZone].Attachment.EndpointId, IgRT)
+	fmt.Println("Successfully route 10.0.0.0/24 to the Firewall Endpoint in IgRT ")
```
fahlmant, line 80: I recommend wrapping this all in an if block or something and getting a command from the cli args of this binary. Like `firewallConfig create`. That way, it will be easy to create `firewallConfig destroy`

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+
+}
+
+func (c Client)CreateVPC (CidrBlock string) (ec2.CreateVpcOutput){
```
fahlmant, line 84: does CidrBlock need to be exported here?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	VPCinput := &ec2.CreateVpcInput{
```
fahlmant, line 85: Does VPCinput and VPCresult need to be exported in this function?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+		CidrBlock: aws.String(CidrBlock),
+	}
+
+	VPCresult, err := c.ec2Client.CreateVpc(VPCinput)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
```
fahlmant, line 92: Are you looking for a specific AWS error code here? If not, this isn't gaining anything.

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+			default:
+				fmt.Println(aerr.Error())
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+
+	fmt.Println("Successfully created a vpc")
```
fahlmant, line 101:
```suggestion
	fmt.Println("Successfully created VPC <vpc id here> with name <name>")
```

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	//Enable DNSHostname
+	DNSHostnameinput := &ec2.ModifyVpcAttributeInput{
```
fahlmant, line 103: Does this need to be exported?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+		EnableDnsHostnames: &ec2.AttributeBooleanValue{
+			Value: aws.Bool(true),
+		},
+		VpcId: aws.String(*VPCresult.Vpc.VpcId),
+	}
+
+	_, err = c.ec2Client.ModifyVpcAttribute(DNSHostnameinput)
```
fahlmant, line 110: 1. Please handle this error
2. Can you add a log here if this is successful?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+
+	return *Subnet
+}
+
+func (c Client) CreateRouteTable (Vpc ec2.CreateVpcOutput, Subnet ec2.CreateSubnetOutput)(ec2.CreateRouteTableOutput){
```
fahlmant, line 170: Should this be called `createRouteTableForSubnet?`

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+				fmt.Println(aerr.Error())
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+	return *RT
+}
+
+func (c Client) CreateRouteTableIG (Vpc ec2.CreateVpcOutput, IG ec2.CreateInternetGatewayOutput)(ec2.CreateRouteTableOutput){
```
fahlmant, line 204: `createRouteTableForIGW`?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+	time.Sleep(wait)
+	fmt.Println("Successfully create a NAT Gateway")
+	return *NGresult
+}
+func (c Client) CreateRoute(CidrBlock string, GatewayID string, RouteTable ec2.CreateRouteTableOutput){
```
fahlmant, line 273: `createRouteWithGateway`?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+			fmt.Println(err.Error())
+		}
+	}
+	fmt.Println("Successfully created a Firewall. Having to wait a long time for the endpoint to be ready!")
+	time.Sleep(wait2)
+
+	return *Firewall
+
+}
+func (c Client) DescribeFirewall (Firewall networkfirewall.CreateFirewallOutput)(networkfirewall.DescribeFirewallOutput){
```
fahlmant, line 373: What's the reason to wrap the Describe call in 377 in another function?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,410 @@
...
+	//Create route 0.0.0.0/0 in FirewallSubnet for IG
+	awsClient.CreateRoute("0.0.0.0/0", *IG.InternetGateway.InternetGatewayId, FirewallRT)
+	fmt.Println("Successfully Created a route 0.0.0.0/0 to IGW in Firewall Subnet")
+
+	//Create Firewall
+	Firewall := awsClient.CreateFirewall(FirewallSubnet, Vpc)
+	//Describe Firewall
+	DescribeFirewall :=awsClient.DescribeFirewall(Firewall)
+	//Create route 0.0.0.0/0 in PublicRT for FirewallEndpoint
+	awsClient.CreateRouteToFirewall("0.0.0.0/0", *DescribeFirewall.FirewallStatus.SyncStates[*FirewallSubnet.Subnet.AvailabilityZone].Attachment.EndpointId, PublicRT)
```
fahlmant, line 76:
```suggestion
        firewallEndpointId := DescribeFirewall.FirewallStatus.SyncStates[*FirewallSubnet.Subnet.AvailabilityZone].Attachment.EndpointId

	awsClient.CreateRouteToFirewall("0.0.0.0/0",firewallEndpointId, PublicRT)
```

fahlmant (changes requested):
Overall looks really good :)
I'm seeing the same pattern:
- Functions are all exported, do they need to be?
- Variable are all exported, do they need to be?
- All errors are unwrapped to AWS errors but then printed the same as they would otherwise. Do they need to be unwrapped?

pkg/awsconfig/testing/firewallConfig.go
bng0y: should we look into polling?

fahlmant (commented):
Looking again, there's also some missing error handling. So far all errors just print out, but what if the VPC fails to create? Currently, the script will keep running and try to create everything else.

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+
+var wait2 = 360 * time.Second
+
+
+func main() {
+	var ec2Client ec2iface.EC2API
+	var firewallClient networkfirewalliface.NetworkFirewallAPI
+
+	region := os.Getenv("REGION")
+	profile := os.Getenv("PROFILE")
```
fahlmant, line 33: Thought about supporting command line args here?

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+		if _, err := sess.Config.Credentials.Get(); err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case "NoCredentialProviders":
+					fmt.Println("could not create AWS session: ", err)
+				default:
+					fmt.Println("could not create AWS session: ", err)
+				}
+			}
+		}
```
fahlmant, line 64: IMO this can be a function

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+
+		ec2Client = ec2.New(sess)
+		//create firewall client
+		firewallClient = networkfirewall.New(sess)
+	}
+
+	awsClient := Client{
+		ec2Client,
+		firewallClient,
+	}
```
fahlmant, line 74: I this can all be combined into one statement

fahlmant (changes requested):
I'm still seeing no errors returned from any of the calls, as well as separating out AWS errors without explicit handling.

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+			fmt.Println(err.Error())
+		}
+	}
+	fmt.Println("Successfully enabled DNSHostname for the newly created VPC")
+	return *VPC
+}
+
+func (c Client) CreateInternetGateway(Vpc ec2.CreateVpcOutput) ec2.CreateInternetGatewayOutput {
+	IGinput := &ec2.CreateInternetGatewayInput{}
+	IGresult, err := c.ec2Client.CreateInternetGateway(IGinput)
```
fahlmant, line 157:
```suggestion
	IGresult, err := c.ec2Client.CreateInternetGateway(&ec2.CreateInternetGatewayInput{})
```

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+		}
+	}
+
+	fmt.Println("Successfully created IG")
+	//Attach the InternetGateway to the VPC
+	IGAttachinput := &ec2.AttachInternetGatewayInput{
+		InternetGatewayId: aws.String(*IGresult.InternetGateway.InternetGatewayId),
+		VpcId:             aws.String(*Vpc.Vpc.VpcId),
+	}
+	_, err = c.ec2Client.AttachInternetGateway(IGAttachinput)
```
fahlmant, line 175:
```suggestion
	IGAttachinput := 
	_, err = c.ec2Client.AttachInternetGateway(&ec2.AttachInternetGatewayInput{
		InternetGatewayId: aws.String(*IGresult.InternetGateway.InternetGatewayId),
		VpcId:             aws.String(*Vpc.Vpc.VpcId),
	}
	)
```

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+	fmt.Println("Successfully enabled DNSHostname for the newly created VPC")
+	return *VPC
+}
+
+func (c Client) CreateInternetGateway(Vpc ec2.CreateVpcOutput) ec2.CreateInternetGatewayOutput {
```
fahlmant, line 155: Given this function is attaching an IGW to a VPC, maybe the name should be `CreateInternetGatewayForVPC` and should just take a VPC ID as input.

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+	fmt.Println("Successfully attached IG to VPC")
+
+	return *IGresult
+}
+
+func (c Client) CreateSubnet(CidrBlock string, Vpc ec2.CreateVpcOutput) ec2.CreateSubnetOutput {
```
fahlmant, line 191:
```suggestion
func (c Client) CreateSubnet(CidrBlock string, vpcID string) ec2.CreateSubnetOutput {
```

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+			}
+		} else {
+			fmt.Println(err.Error())
+		}
+	}
+
+	return *Subnet
+}
+
+func (c Client) CreateRouteTableForSubnet(Vpc ec2.CreateVpcOutput, Subnet ec2.CreateSubnetOutput) ec2.CreateRouteTableOutput {
```
fahlmant, line 211:
```suggestion
func (c Client) CreateRouteTableForSubnet(vpcID string, subnetID string) ec2.CreateRouteTableOutput {
```

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+
+	return *IGresult
+}
+
+func (c Client) CreateSubnet(CidrBlock string, Vpc ec2.CreateVpcOutput) ec2.CreateSubnetOutput {
+	Subnetinput := &ec2.CreateSubnetInput{
+		CidrBlock: aws.String(CidrBlock),
+		VpcId:     aws.String(*Vpc.Vpc.VpcId),
+	}
+	Subnet, err := c.ec2Client.CreateSubnet(Subnetinput)
```
fahlmant, line 196:
```suggestion
	
	Subnet, err := c.ec2Client.CreateSubnet(Subnetinput := &ec2.CreateSubnetInput{
		CidrBlock: aws.String(CidrBlock),
		VpcId:     aws.String(*Vpc.Vpc.VpcId),
	})
```

pkg/awsconfig/testing/firewallConfig.go
```diff
@@ -0,0 +1,479 @@
...
+
+	return *Subnet
+}
+
+func (c Client) CreateRouteTableForSubnet(Vpc ec2.CreateVpcOutput, Subnet ec2.CreateSubnetOutput) ec2.CreateRouteTableOutput {
+	RouteTable1input := &ec2.CreateRouteTableInput{
+		VpcId: aws.String(*Vpc.Vpc.VpcId),
+	}
+
+	RT, err := c.ec2Client.CreateRouteTable(RouteTable1input)
```
fahlmant, line 216:
```suggestion
	RouteTable1input := 

	RT, err := c.ec2Client.CreateRouteTable(&ec2.CreateRouteTableInput{
		VpcId: aws.String(*Vpc.Vpc.VpcId),
	})
```

pkg/awsconfig/testing/firewallConfig.go
lnguyen1401: well since the amount of argument is different ( passing PROFILE vs passing ENV) i think it might be safer like this than command lines args

fahlmant:
```
AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=bar AWS_SESSION_TOKEN=baz  ./firewall-create

AWS_PROFILE=devaccount AWS_REGION=us-west-2 ./firewall-create

firewall-create --profile devaccount --region us-east-1
```

All three of these should work

fahlmant (changes requested):
A few things still remain
1. The errors are not being returned from the functions and handled. This means the script will keep going even if one of the previous required steps fails.
2. The errors are being parsed into AWS errors, but nothing is being done with them different than  non-AWS errors. If the AWS errors should have different logic, they can be handled separately. Otherwise, there's no reason to handle them differently. 
3. The Profile and Env client functions can probably be mostly combined into one function. 
4. Region should also be supported via cli option (i.e. --region)

pkg/awsconfig/testing/firewallConfig.go
lnguyen1401: removed

pkg/awsconfig/testing/firewallConfig.go
lnguyen1401: implemented -p $profile -r $region

pkg/awsconfig/testing/firewallConfig.go
lnguyen1401: implemented

pkg/awsconfig/testing/firewallConfig.go
lnguyen1401: implemented

PR #62, opened 2022-04-27
Parallelize egress endpoint checks 

abyrne55:
Egress endpoint checks are embarrassingly parallelizable, so we should parallelize them. This PR does so by running `ValidateReachability()` in a goroutine, communicating errors (i.e., unreachable endpoints) via a `failures` channel. We also explicitly pass `host` and `port` into the anonymous goroutine to avoid out-of-order memory access issues.

This addresses [OSD-10647](https://issues.redhat.com/browse/OSD-10647), where a lack of parallelization caused the verifier to run for over 30 seconds. With this PR, the verifier binary (inside the docker image, not including instance creation time) takes less than 3 seconds to run.

### DIFF
--- a/build/bin/network-validator.go
+++ b/build/bin/network-validator.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"sync"
 	"time"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


boranx (approved):
Hi @abyrne55 
This is really fast :) Nice improvement :+1: 
Given that all the validations are running simultaneously, we can think about loosening the timeout a bit more (I got some errors on 1s but worked well under 2s)

PR #77, opened 2022-05-06
Removing cert-api.access.redhat.com from validator config

fahlmant:
This url is not listed in the documentation as a required egress point [1], and causes issues with PR #50. Removing as its not required. 

[1] https://deploy-preview-32552--osdocs.netlify.app/openshift-rosa/latest/rosa_getting_started/rosa-aws-prereqs.html#osd-aws-privatelink-firewall-prerequisites

### DIFF
--- a/build/config/config.yaml
+++ b/build/config/config.yaml
@@ -22,9 +22,6 @@ endpoints:
   - host: quay-registry.s3.amazonaws.com
     ports:
       - 443
-  - host: cert-api.access.redhat.com
-    ports:
-      - 443
   - host: api.access.redhat.com
     ports:
       - 443


ritmun:
Just curious: since telemetry requires these URLs, wouldn't we need to verify those if customer wishes to NOT disable telemetry?

fahlmant:
Is there documentation on where this is required by telemetry? The official docs I linked are what customers need to follow so the egress test should only check those. If this URL is required, we need to ask the docs team to add it.

fahlmant:
Nevermind, I was looking at an old preview :face_exhaling:

fahlmant:
https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/rosa-aws-prereqs.html
has it, closing PR

PR #78, opened 2022-05-09
OSD-11009: make test time out error pre-defined instead of generic

anispate:
The purpose of this PR was to move the error definition which is mentioned in the helpers.go to errors.go file so, that we can throw a test time out error instead of generic(unhandled) error

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -1,7 +1,6 @@
 package errors
 
 import (
-	_ "embed"
 	"errors"
 	"fmt"
 )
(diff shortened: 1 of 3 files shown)


boranx:
@anispate thanks for your refactoring! Looks like there's also a binary pushed into PR `pkg/awsconfig/testing/firewallConfig`
Can you exclude this so the PR only would contain your changes?

pkg/errors/errors.go
```diff
@@ -1,11 +1,19 @@
 package errors
 
-import "fmt"
+import (
+	_ "embed"
+	"errors"
+	"fmt"
```
fahlmant, line 6: Is this used in this package?

PR #80, opened 2022-05-09
OSD-9234 - Adding a new make target to validate formating

bdematte:
Making a new make target to be called by the new step in CI


bdematte:
/hold 
Just saw a regression in the Makefile

boranx (approved):
/lgtm
Thanks!

PR #93, opened 2022-05-16
organize readmes and examples

ritmun:
Organized readme and examples

docs go under docs/
examples go under examples/

Main readme doc leads to examples as follows: 

Main readme -> AWS specific readme -> examples and example readmes


examples/aws/firewall/firewallConfig.go
```diff
@@ -1,4 +1,4 @@
-package main
+package firewall
```
fahlmant, line 1: Go programs require a main package. This change breaks the script

boranx: I think in general we shouldn't propose changes in source code unless we're sure/executed in local and see it's working as expected

ritmun: that was my IDE's brilliant suggestion :(

ritmun: reverted and tested script locally

fahlmant (changes requested):
The build is failing, looks like go.mod and go.sum are required

examples/aws/firewall/firewallConfig.go
ritmun: @boranx I agree with that. Also that every piece of code should have a test in the pipeline pre-merge to make sure nothing like that skips our attention.

ritmun:
> The build is failing, looks like go.mod and go.sum are required

thanks, fixed

boranx:
@ritmun could you rebase this so we can merge?

ritmun:
> @ritmun could you rebase this so we can merge?

rebased

PR #94, opened 2022-05-16
updated credentials setup to match SOP updates

ritmun:
credential setup readme is updated to match review of OPS-SOP changes

### DIFF
--- a/AWS.md
+++ b/AWS.md
@@ -15,8 +15,11 @@
 
 ## Setup ##
 ### AWS Environment ###
-Set up your environment to use the correct credentials for the AWS account for the target cluster. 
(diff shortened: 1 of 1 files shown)


fahlmant (changes requested):
This repo is public and should be general to anyone wanting to use this code base. The linked guide is private, and specific to Red Hat SREP setup. Anyone with a valid AWS credential set can use this repo, so I think keeping the instructions non-specific here is important.

ritmun:
> This repo is public and should be general to anyone wanting to use this code base. The linked guide is private, and specific to Red Hat SREP setup. Anyone with a valid AWS credential set can use this repo, so I think keeping the instructions non-specific here is important.

agreed. closing

PR #97, opened 2022-05-19
OSD-11306 - Update dockerfile to have a working entrypoint 

bdematte:
The entrypoint defined in the Dockerfile needs to be in the PATH to be able to behave properly.


boranx (commented):
/lgtm

PR #100, opened 2022-05-26
docs: remove duplicate terminateinstance, add create tags

boranx:


AWS.md
```diff
@@ -51,13 +51,13 @@ Ensure that the AWS credentials being used have the following permissions.
     {
       "Effect": "Allow",
       "Action": [
+        "ec2:CreateTags",
```
boranx, line 54: https://github.com/openshift/osd-network-verifier/blob/main/pkg/cloudclient/aws/private.go#L184
This creates the tag unless it's created already. Therefore, it needs permission... It's also possible to narrow the context: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/supported-iam-actions-tagging.html

PR #103, opened 2022-06-03
OSD-11859 - Removing SFTP as not Mandatory and removing dupplicated check

bdematte:
As part of the issues faced by in Prod when we activated osd-network-verifier, we found that the SFTP access was checked (and blocking is failed) while it is only Recommended in the docs. To align with docs, removing the check.
Reference : [OSD-11859](https://issues.redhat.com/browse/OSD-11859)
In addition, removing dupplicated check for quay-registry.s3.amazonaws.com

### DIFF
--- a/build/config/config.yaml
+++ b/build/config/config.yaml
@@ -38,9 +38,6 @@ endpoints:
   - host: storage.googleapis.com
     ports:
       - 443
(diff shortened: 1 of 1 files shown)


fahlmant (approved):
/approve
/lgtm
@bdematte Don't forget to bump the tag ID after this merges

PR #104, opened 2022-06-03
OSD-11912 - Remove api.pagerduty.com from checked domains

bdematte:
api.pagerduty.com is used only for service/account management, and not by the monitoring tools (which are using events.pagerduty.com) so removing it from the domains to be validated. 
Reference : [OSD-11912](https://issues.redhat.com/browse/OSD-11912)


fahlmant (approved):
/approve

PR #107, opened 2022-06-07
 osdctl-11717 , osd-11580 add profile flag, refactor client input to struct

ritmun:
accommodation for profile flag in egress command (linked to osdctl implementation of verify egress command osd-11717) 

jira https://issues.redhat.com/browse/OSD-11717 

to test locally: 

make build
./osd-network-verifier egress --profile your_dev_profile --subnet-id subnet_id  --region region


pkg/cloudclient/aws/private.go
```diff
@@ -124,9 +176,9 @@ func (c *Client) validateInstanceType(ctx context.Context) error {
 		// Check for invalid instance type error and return a cleaner error
 		re := regexp.MustCompile("400.*api error InvalidInstanceType")
 		if re.Match([]byte(err.Error())) {
-			err = fmt.Errorf("Instance type %s does not exist", c.instanceType)
+			err = fmt.Errorf("instance type %s does not exist", c.instanceType)
```
ritmun, line 179: Changed initial to lowercase as my local lint was showing warnings

pkg/cloudclient/aws/private.go
```diff
@@ -62,15 +62,25 @@ var (
...
-func newClient(ctx context.Context, logger ocmlog.Logger, accessID, accessSecret, sessiontoken, region, instanceType string, tags map[string]string) (*Client, error) {
-	cfg, err := config.LoadDefaultConfig(ctx,
-		config.WithRegion(region),
-		config.WithCredentialsProvider(credentials.StaticCredentialsProvider{
-			Value: aws.Credentials{
-				AccessKeyID: accessID, SecretAccessKey: accessSecret, SessionToken: sessiontoken,
-			},
-		}),
-	)
+func newClient(ctx context.Context, logger ocmlog.Logger, accessID, accessSecret, sessiontoken, region,
```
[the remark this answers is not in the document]
ritmun: updated

anispate:
Summary:
All tests pass for the ONV
Success

cmd/dns/cmd.go
```diff
@@ -49,9 +49,22 @@ func NewCmdValidateDns() *cobra.Command {
...
+					logger.Info(ctx, "Using provided AWS credentials")
+				}
+				// The use of t3.micro here is arbitrary; we just need to provide any valid machine type
+				cli, err = cloudclient.NewClient(ctx, logger, config.region, "t3.micro", nil, "aws", config.awsProfile)
+
+			} else {
+				//	todo after GCP is implemented, check GCP type using creds
+				logger.Info(ctx, "GCP cloud credentials found.")
+				cli, err = cloudclient.NewClient(ctx, logger, config.region, "", nil, "gcp", config.awsProfile)
+			}
```
fahlmant, line 67: This should be removed until we're ready to handle GCP credentials. Currently, this leads to the binary incorrectly sayings it using GCP credentials when its not, and passing in config.awsProfile, which is confusing.

cmd/egress/cmd.go
```diff
@@ -99,3 +97,21 @@ are set correctly before execution.
...
+		} else {
+			logger.Info(ctx, "Using provided AWS credentials")
+		}
+		return cloudclient.NewClient(ctx, logger, config.region, config.instanceType, config.cloudTags, "aws",
+			config.awsProfile)
+
+	} else {
+		//	todo after GCP is implemented, check GCP type using creds
+		logger.Info(ctx, "GCP cloud credentials found.")
+		return cloudclient.NewClient(ctx, logger, config.region, config.instanceType, config.cloudTags, "gcp", "")
```
fahlmant, line 115: Same thought here as mentioned above

cmd/egress/cmd.go
```diff
@@ -99,3 +97,21 @@ are set correctly before execution.
 	return validateEgressCmd
 
 }
+
+func NewClient(ctx context.Context, logger ocmlog.Logger, config egressConfig) (cloudclient.CloudClient, error) {
```
fahlmant, line 101: Why is the egress pkg implementing its own NewClient function? Shouldn't we make cloudclient handle this instead?

fahlmant (changes requested):
We're adding more and more AWS specific configs to the commands themselves, rather than implementing logic in our cloudclient pkg to handle the cloud stuff behind the scenes. If possible, lets move as much as we can to the cloudclient, and maybe add `--aws` and `--gcp` flags to the commands to help determine which option set to use. Then we can have `awsOptions` and `gcpOptions`, just as a first thought.

ritmun:
> We're adding more and more AWS specific configs to the commands themselves, rather than implementing logic in our cloudclient pkg to handle the cloud stuff behind the scenes. If possible, lets move as much as we can to the cloudclient, and maybe add `--aws` and `--gcp` flags to the commands to help determine which option set to use. Then we can have `awsOptions` and `gcpOptions`, just as a first thought.

That's a good idea. Should we add this to the done criteria of this card? https://issues.redhat.com/browse/OSD-11384 Seems related..

cmd/egress/cmd.go
ritmun: It does handle the actual NewClient function. We can rename this to NewClientCaller or something. I separated this from the cmd following our previous comments on osdctl where I removed extra logic to a separate function to keep the cmd "Run" param simple. I can put it back up in "Run".. No preference.

cmd/dns/cmd.go
ritmun: updated

cmd/egress/cmd.go
ritmun: updated

cmd/egress/cmd.go
ritmun: Forgot to add, I didn't do the same for dns and byovpc cmds, because I wasn't working on those 2 so didn't want to touch them too much other than preventing errors. Moving it back so they should all be the same.

examples/aws/verify_egressv2.go
```diff
@@ -3,33 +3,24 @@ package aws
...
 	"fmt"
-	"os"
 	"time"
 
-	"github.com/aws/aws-sdk-go-v2/credentials"
 	ocmlog "github.com/openshift-online/ocm-sdk-go/logging"
 	"github.com/openshift/osd-network-verifier/pkg/cloudclient"
 )
 
 func extendValidateEgressV2() {
```
boranx, line 12: so, I can see we no longer need v1/v2 differentation? https://github.com/openshift/osd-network-verifier/blob/main/docs/aws/aws.md#112-go-implementation-examples

Is this change suitable w/ ocm-backend? I think it is worth exploring/ pinging them to ask we're changing the way of client generation.
 https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/aws/inflight_service.go#L9
&
 https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/aws/inflight_service.go#L229

pkg/cloudclient/aws/aws.go
```diff
@@ -51,40 +61,18 @@ func (c *Client) VerifyDns(ctx context.Context, vpcID string) *output.Output {
...
+		return nil, fmt.Errorf("unable to create AWS client: %w", err)
 	}
+	return
+}
 
+func GetEc2ClientFromInput(input *ClientInput) (ec2.Client, error) {
+	ec2Client, err := getEc2ClientFromInput(*input)
 	if err != nil {
-		return nil, fmt.Errorf("Unable to create AWS client: %w", err)
+		return ec2Client, fmt.Errorf("unable to create EC2 Client: %w", err)
```
boranx, line 75: return (nil, err) if there's a problem with getting. I think this is a general convention

pkg/cloudclient/aws/private.go
```diff
@@ -62,35 +64,85 @@ var (
...
+			config.WithSharedConfigProfile(input.Profile),
+			config.WithRegion(input.Region),
+		)
+		if err != nil {
+			return ec2Client, fmt.Errorf("can not get AWS config from profile `%s`", input.Profile)
+		}
+
+	} else {
+		//Get config from access keys if provided
+		cred := credentials.NewStaticCredentialsProvider(input.AccessKeyId,
```
boranx, line 83:
```go
		//Get config from access keys if provided
		cred := credentials.NewStaticCredentialsProvider(input.AccessKeyId,
			input.SecretAccessKey, input.SessionToken)
		switch cr := interface{}(cred).(type) {
```
are you sure the above is looking right? you're getting `credentials.NewStaticCredentialsProvider` from "github.com/aws/aws-sdk-go-v2/credentials", so your switch-case would always catch `case awscredsv2.StaticCredentialsProvider`

boranx (changes requested):
hey @ritmun,
thanks for the PR!

I've left some questions that need addressing

pkg/cloudclient/aws/aws.go
ritmun: Unfortunately we can't return nil for ec2client. We get error: "Cannot use 'nil' as the type ec2.Client".  Do you know if there;s a workaround for that?

examples/aws/verify_egressv2.go
ritmun: `Is this change suitable w/ ocm-backend?` :  Another great point. I updated the function signature to allow modifying args list without breaking dependencies in future. I'll contact ocm backend team to update their code once we merge this, and before moving to prod. No point contacting people until our change is finalized.

fahlmant:
@ritmun The fmt job is failing

pkg/cloudclient/aws/private.go
boranx: in `cmd/`, we fixed it into v2.
But in `pkg/`, NewClient() supports both v1/v2: https://github.com/openshift/osd-network-verifier/blob/main/pkg/cloudclient/aws/aws.go#L54 (as `creds interface{}`)

ritmun:
I've converted input to client into a struct. I've used context to pass some of the values in order to avoid cyclic imports to and from cloudclient <> egress-cmd.  This is working in the local tests I run but any ideas for better ways for polymorphism of this kind are welcome.

cmd/egress/cmd.go
```diff
@@ -6,29 +6,17 @@ import (
...
 var (
-	defaultTags            = map[string]string{"osd-network-verifier": "owned", "red-hat-managed": "true", "Name": "osd-network-verifier"}
-	regionEnvVarStr string = "AWS_REGION"
-	regionDefault   string = "us-east-2"
+	defaultTags     = map[string]string{"osd-network-verifier": "owned", "red-hat-managed": "true", "Name": "osd-network-verifier"}
+	regionEnvVarStr = "AWS_REGION"
+	regionDefault   = "us-east-2"
 )
 
-type egressConfig struct {
```
ritmun, line 21: This struct is defined in cloudclient and fields required by AWS are added to AWS client to remove need for cyclic imports

docs/aws/aws.md
```diff
@@ -20,20 +20,21 @@
...
-  For STS credentials, also:
-    ```shell 
-    export AWS_SESSION_TOKEN=<YOUR_SESSION_TOKEN_STRING> 
-    ```
-  Export any other AWS environment vars:
-    ```shell
-    export AWS_REGION=<VPC_AWS_REGION>
-    ````
+- Obtain a valid set of AWS secret and key for the target account and use them in one of the following ways:
+  - Set them as an AWS profile in you ~/.aws/credentials file as prescribed in [this AWS doc.](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_profiles.html)
```
ritmun, line 24: note for using profile flag

pkg/cloudclient/aws/private.go
ritmun: We can remove v1 completely, but I've didn't want to introduce that change in this PR. Can we card it out?

cmd/egress/cmd.go
```diff
@@ -82,20 +70,21 @@ are set correctly before execution.
...
-	validateEgressCmd.Flags().StringVar(&config.instanceType, "instance-type", "t3.micro", "(optional) compute instance type")
-	validateEgressCmd.Flags().StringVar(&config.region, "region", getDefaultRegion(), fmt.Sprintf("(optional) compute instance region. If absent, environment var %[1]v will be used, if set", regionEnvVarStr, regionDefault))
-	validateEgressCmd.Flags().StringToStringVar(&config.cloudTags, "cloud-tags", defaultTags, "(optional) comma-seperated list of tags to assign to cloud resources e.g. --cloud-tags key1=value1,key2=value2")
-	validateEgressCmd.Flags().BoolVar(&config.debug, "debug", false, "(optional) if true, enable additional debug-level logging")
-	validateEgressCmd.Flags().DurationVar(&config.timeout, "timeout", 2*time.Second, "(optional) timeout for individual egress verification requests")
-	validateEgressCmd.Flags().StringVar(&config.kmsKeyID, "kms-key-id", "", "(optional) ID of KMS key used to encrypt root volumes of compute instances. Defaults to cloud account default key")
+	// egress test flags
+	validateEgressCmd.Flags().StringVar(&config.VpcSubnetID, "subnet-id", "", "source subnet ID")
```
ritmun, line 74: There should be no diff here other than added line for profile flag, one correction for "instance-type" line and use for published "config" vars instead of private ones.

pkg/cloudclient/aws/private.go
ritmun: @boranx on a second look, I see what you were suggesting. I removed the switch loop. I don't see a need for it any more. Downstream can pass in creds and onv can create ec2client object regardless of version.

docs/aws/aws.md
```diff
@@ -103,29 +104,30 @@ repeat the verification process for each subnet ID.
...
+         ```
+   
        Get cli help:
     
         ```shell
         ./osd-network-verifier egress --help
         ```
 
-##### 1.1.2 Go implementation Examples #####
-- [AWS Go SDK v1](../../examples/aws/verify_egressv1.go)
```
ritmun, line 127: There's no need to pass version specific creds object to client any more. Downstream was previously creating version specific cred object, but will now be providing either version agnostic profile or AWS keys to cloudclient.

pkg/cloudclient/aws/aws.go
boranx: Hey @ritmun 

you can return `*ec2.Client` instead in the function. 
Also, I think we can omit the creation of ec2client in the beginning as `ec2.NewFromConfig(cfg)` can create one.

(e.g 
```go
func getEc2ClientFromInput(input ClientInput) (*ec2.Client, error) {
	//Get config from profile if provided
	if input.Profile != "" {
		cfg, err := config.LoadDefaultConfig(input.Ctx,
			config.WithSharedConfigProfile(input.Profile),
			config.WithRegion(input.Region),
		)

		return ec2.NewFromConfig(cfg), err
	}

	cfg, err := config.LoadDefaultConfig(input.Ctx,
		config.WithRegion(input.Region),
		config.WithCredentialsProvider(credentials.StaticCredentialsProvider{
			Value: aws.Credentials{
				AccessKeyID: input.AccessKeyId, SecretAccessKey: input.SecretAccessKey,
				SessionToken: input.SessionToken,
			},
		}),
	)

	return ec2.NewFromConfig(cfg), err
}
```
)

pkg/cloudclient/aws/private.go
```diff
@@ -62,35 +62,60 @@ var (
...
-func newClient(ctx context.Context, logger ocmlog.Logger, accessID, accessSecret, sessiontoken, region, instanceType string, tags map[string]string) (*Client, error) {
-	cfg, err := config.LoadDefaultConfig(ctx,
-		config.WithRegion(region),
-		config.WithCredentialsProvider(credentials.StaticCredentialsProvider{
-			Value: aws.Credentials{
-				AccessKeyID: accessID, SecretAccessKey: accessSecret, SessionToken: sessiontoken,
-			},
-		}),
-	)
+func getEc2ClientFromInput(input ClientInput) (ec2.Client, error) {
```
boranx, line 65: At first glance, `aws.go/ GetEc2ClientFromInput()` and `private.go/ getEc2ClientFromInput()` has pretty much same logic.
It looks a bit repeating itself so it would be nice if we unified it

pkg/cloudclient/aws/private.go
boranx: ack, that looks cleaner :) I think eventually we'd need a jira card to make sure this is imported to uhc-backend gracefully
(As they no more need creds, instead it's just: `cli, err := cloudclient.NewClient(context.TODO(), logger, cmdOptions)`)

boranx (commented):
nice improvements, @ritmun I've left additional questions but overall looks good

pkg/cloudclient/aws/aws.go
ritmun: This looks good. Updated.

pkg/cloudclient/aws/private.go
ritmun: I'm adding it it as a todo in this card before promoting to prod

cmd/egress/cmd.go
```diff
@@ -6,29 +6,17 @@ import (
...
 	"github.com/spf13/cobra"
 )
 
 var (
-	defaultTags            = map[string]string{"osd-network-verifier": "owned", "red-hat-managed": "true", "Name": "osd-network-verifier"}
-	regionEnvVarStr string = "AWS_REGION"
-	regionDefault   string = "us-east-2"
+	defaultTags     = map[string]string{"osd-network-verifier": "owned", "red-hat-managed": "true", "Name": "osd-network-verifier"}
+	regionEnvVarStr = "AWS_REGION"
+	regionDefault   = "us-east-2"
```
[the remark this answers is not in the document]
ritmun: moved these to cloudclient, also moved time default there

ritmun:
> nice improvements, @ritmun I've left a suggestion for `getEc2ClientFromInput()` but overall looks good

thanks! it's updated.

cmd/egress/cmd.go
ritmun: update: default values should now come from cloudclient.go

pkg/cloudclient/cloudclient.go
```diff
@@ -25,22 +55,52 @@ type CloudClient interface {
...
+	}
+	if options.CloudType == GCP {
+		return GCP
+	}
+	return "unknown"
+}
+
+func NewClient(ctx context.Context, logger ocmlog.Logger,
+	options CmdOptions) (CloudClient, error) {
+	logger.Info(ctx, "Using region: %s", options.Region)
```
fahlmant, line 78: This won't print out anything if profile isn't set on the cli, but the test still runs.

fahlmant: Example:

```
 AWS_ACCESS_KEY_ID=$ACCESS_KEY AWS_SECRET_ACCESS_KEY=$SECRET_KEY ./osd-network-verifier egress --subnet-id $SUBNET_ID  
Using region: 
Using AWS secret key
Created instance with ID: i-06261950f8802ec25
EC2 Instance: i-06261950f8802ec25 Running
Terminating ec2 instance with id i-06261950f8802ec25
Summary:
All tests pass!
Success
```

cmd/dns/cmd.go
```diff
@@ -31,7 +24,7 @@ func getDefaultRegion() string {
 	}
 }
 func NewCmdValidateDns() *cobra.Command {
-	config := dnsConfig{}
+	config := cloudclient.CmdOptions{}
```
fahlmant, line 27: Why was this changed? The DNS test doesn't need all the same cmd options as the egress test. Each test may have its  own subset or superset of options it needs that aren't related to the cloudclient

pkg/cloudclient/cloudclient.go
ritmun: I'd like to remove it

cmd/dns/cmd.go
```diff
@@ -41,23 +34,18 @@ func NewCmdValidateDns() *cobra.Command {
...
 			if err != nil {
 				fmt.Printf("Unable to build logger: %s\n", err.Error())
 				os.Exit(1)
 			}
-
-			logger.Warn(ctx, "Using region: %s", config.region)
-			creds := credentials.NewStaticCredentialsProvider(os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), os.Getenv("AWS_SESSION_TOKEN"))
-			// The use of t3.micro here is arbitrary; we just need to provide any valid machine type
-			cli, err := cloudclient.NewClient(ctx, logger, creds, config.region, "t3.micro", nil)
+			cli, err := cloudclient.NewClient(ctx, logger, config)
```
fahlmant, line 43:
```suggestion
			client, err := cloudclient.NewClient(ctx, logger, config)
```

cmd/egress/cmd.go
fahlmant: This struct should stay local to the egress command and only be used here. The cloudclient should hvae its own struct with only options that apply to creating the client.

ritmun: It's not possible to implement it that way without cycling imports. The imports get deeper and import cycles develop. It's counter intuitive but this is what came out of a few days of trying multiple ways. I'm open to refactor later on.

fahlmant: What cyclic imports were happening?

cmd/dns/cmd.go
ritmun: done

ritmun: done

cmd/egress/cmd.go
fahlmant: I'm -1 to refactoring later. This is a major change to how this api works and will be difficult to unwind.

ritmun: It's not possible. We are aiming to use struct instead of vars to pass to the client, so the struct must be defined in the client. If it is defined in cmd, 2 major things are wrong with that. 
  1. Downstream code has to define the struct, so we are not providing them a streucture for input. 
  2. cloudclient must import egress command and go does not like the cyclic import. We shouldn't be importing downstream code.

pkg/cloudclient/cloudclient.go
```diff
@@ -25,22 +55,54 @@ type CloudClient interface {
...
+	if options.CloudType == GCP {
+		return GCP
+	}
+	return "unknown"
+}
+
+func NewClient(ctx context.Context, logger ocmlog.Logger,
+	options CmdOptions) (CloudClient, error) {
+
+	logger.Info(ctx, "Using region: %s", getDefaultRegion())
```
fahlmant, line 79: This prints `us-east-1` regardless of what region is provided

```
AWS_ACCESS_KEY_ID=$ACCESS_KEY AWS_SECRET_ACCESS_KEY=$SECRET_KEY  ./osd-network-verifier egress --subnet-id $SUBNET_ID  --region us-west-2
Using region: us-east-1
Using AWS secret key
```

cmd/egress/cmd.go
ritmun: I'm with you on -1 to that refactor as it won't work.

fahlmant: The struct for the AWS client should be dfined in the AWS code. The struct to hold options for the specific test should be help in the test definition. There should be two different structs. The code in egress.go should build the struct needed for AWS Client (or GCP client) And pass it to the NewClient function. There should be no cyclic imports there

pkg/cloudclient/gcp/gcp.go
```diff
@@ -25,12 +25,22 @@ type Client struct {
...
 
+type ClientInput struct {
+	Ctx          context.Context
+	Logger       ocmlog.Logger
+	Creds        *google.Credentials
+	Region       string
+	InstanceType string
+	Tags         map[string]string
+	Timeout      time.Duration
+}
```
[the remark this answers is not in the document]
ritmun: This choice bothered me as well, for a few days and I tried everything I could find to avoid this.

pkg/cloudclient/cloudclient.go
ritmun: updated

cmd/egress/cmd.go
ritmun: Tried that, still had issues..

pkg/cloudclient/cloudclient.go
```diff
@@ -3,18 +3,48 @@ package cloudclient
...
+	Timeout      time.Duration
+	KmsKeyID     string
+}
 
-	"golang.org/x/oauth2/google"
+var (
+	DefaultTags     = map[string]string{"osd-network-verifier": "owned", "red-hat-managed": "true", "Name": "osd-network-verifier"}
+	RegionEnvVarStr = "AWS_REGION"
+	RegionDefault   = "us-east-1"
+	DefaultTime     = 2 * time.Second
```
fahlmant, line 35: These are AWS specific, and likely won't work for GCP

cmd/egress/cmd.go
fahlmant: What issues specifically?

rafael-azevedo (commented):
@ritmun I think the PR has become a bit bloated since it not only is adding a new feature but doing additional refactors. Its easier to review and merge changes if the PRs are as small as possible for the features. That way we can spend time discussing the merits of a refactor without blocking feature merging. 

At the end of the day on this PR I think its best not to couple all config parameters to on package but to keep them within the packages that use them directly. Even if it might be a better technical solution to couple things in terms of readability and maintainability keeping these structs closer to their source packages is easier.

pkg/cloudclient/cloudclient.go
ritmun: Those are optional. But that's as intended. There's a card which I'm working on - to separate AWS and GCP functionality and make it agnostic. We should go deeper on that card, imo.

rafael-azevedo:
> What
> 
> > @ritmun I think the PR has become a bit bloated since it not only is adding a new feature but doing additional refactors. Its easier to review and merge changes if the PRs are as small as possible for the features. That way we can spend time discussing the merits of a refactor without blocking feature merging.
> > At the end of the day on this PR I think its best not to couple all config parameters to on package but to keep them within the packages that use them directly. Even if it might be a better technical solution to couple things in terms of readability and maintainability keeping these structs closer to their source packages is easier.
> 
> Yes, I worked on this PR for a while and put a lot of effort in it. It was a part of our previous sprint, which is why I'm still following up steadily. I will in future isolate changes, but this PR is near completion. What would you like me to do with it?

I am not the maintainer. I would defer to asking them what to do in this particular PR. In my opinion I would pull the existing work  from adding the profile flag in a new PR without any of the refactor other packages. I'd pass everything in explicitly as is done right now.  

I do see need for some refactoring consolidating some of the multiple parameters passed into some function that are starting to pile up into a input type struct, but maybe in a separate PR.

ritmun:
> > What
> > > @ritmun I think the PR has become a bit bloated since it not only is adding a new feature but doing additional refactors. Its easier to review and merge changes if the PRs are as small as possible for the features. That way we can spend time discussing the merits of a refactor without blocking feature merging.
> > > At the end of the day on this PR I think its best not to couple all config parameters to on package but to keep them within the packages that use them directly. Even if it might be a better technical solution to couple things in terms of readability and maintainability keeping these structs closer to their source packages is easier.
> > 
> > 
> > Yes, I worked on this PR for a while and put a lot of effort in it. It was a part of our previous sprint, which is why I'm still following up steadily. I will in future isolate changes, but this PR is near completion. What would you like me to do with it?
> 
> I am not the maintainer. I would defer to asking them what to do in this particular PR. In my opinion I would pull the existing work from adding the profile flag in a new PR without any of the refactor other packages. I'd pass everything in explicitly as is done right now.
> 
> I do see need for some refactoring consolidating some of the multiple parameters passed into some function that are starting to pile up into a input type struct, but maybe in a separate PR.

Rafael, I appreciate your experience and knowledge and your comment. I've been working with the maintainers of this repo the whole time, and we together have been working on this PR, starting from grooming of this card to each new change added into it, for a while now. I'll proceed with this PR as appropriate. Thanks.

pkg/cloudclient/cloudclient.go
fahlmant: If we're adding these and they're AWS specific, then they should exist in the AWS specific part of the cloudclient.

pkg/cloudclient/cloudclient.go
fahlmant: I don't see a commit pushed to address this. Possible you forgot to push?

fahlmant (changes requested):
The structs for options specific to commands and the structs to hold the cloud client configurations should be separate. osdctl has examples where individual commands have option structs that are separate from configuration for the aws client, that may be a good place to reference. 

Example:
https://github.com/openshift/osdctl/blob/master/cmd/account/console.go#L41

ritmun:
> The structs for options specific to commands and the structs to hold the cloud client configurations should be separate. osdctl has examples where individual commands have option structs that are separate from configuration for the aws client, that may be a good place to reference.
> 
> Examples: https://github.com/openshift/osdctl/blob/master/cmd/account/console.go#L41 https://github.com/openshift/osdctl/blob/master/cmd/account/cli.go#L38

ok. I can do this. I'll have to put this struct as a separate import apart from aws or cloudclient package to avoid import dependencies. our imports are aws > cloudclient > downstream. If we don't want the struct as 

> The structs for options specific to commands and the structs to hold the cloud client configurations should be separate. osdctl has examples where individual commands have option structs that are separate from configuration for the aws client, that may be a good place to reference.
> 
> Examples: https://github.com/openshift/osdctl/blob/master/cmd/account/console.go#L41 https://github.com/openshift/osdctl/blob/master/cmd/account/cli.go#L38

I'd like to implement elements of these examples applicable in this client. I'll take some time to experiment with these. Will re-request review then.

ritmun:
> The structs for options specific to commands and the structs to hold the cloud client configurations should be separate. osdctl has examples where individual commands have option structs that are separate from configuration for the aws client, that may be a good place to reference.
> 
> Examples: https://github.com/openshift/osdctl/blob/master/cmd/account/console.go#L41 https://github.com/openshift/osdctl/blob/master/cmd/account/cli.go#L38

 
	@fahlmant I appreciate you taking time to review this important refactor, and thanks for pointing out some really good catches. An update on this: I just started to try the osdctl example out. I'm not sure this vision of structs applies to our implementation. My conclusion is what we have now is an optimal way of dealing with this design challenge. It can always be improved, but it is sensible for its purpose. 
 
	Let me explain why I say that. On a higher level, our implementation is different than osdctl because onv client is open to downstream cli integration, so we have more constraints on how pluggable we want our client to be. Also, onv has really good polymorphism in both cloud client type gcp/aws and subcommand types. This is also its challenge.  
	
 	On a lower level, on this card, this is the case: I separated egress test specific flags and client flags. Defined the test specific flags in a separate struct under our package so that downstream can use it.   I passed them as 2 separate structs to the client.  This means that each test similarly will be passing its own type of struct to the client and we end up with the same old problem of variable input to the client. The input should be a predefined struct or structs for a stable integration with downstream. With this design, if we add more tests, the getter's signature will change, since each test has its own struct. Also, the client must be agnostic to both cloud types and test types. All it should do is create a cloudclient based on given cli options.  The type of cloud (GCP/AWS) can be easily identified based on values provided and/or “cloudType” flag, however, we can’t/shouldn’t differentiate which test is creating the client based on the input.  
 	
 	I’ve also previously tried to define all these vars in the AWS and GCP clients but then we have a cyclin import issue, as cloudclient imports AWS and downstream imports cloudclient. I didn’t write down why but that option didn’t go anywhere.
 	
  	This is why cloudclient interface should define all variables within it, and those which don’t apply must be left nil. The client should not care which test is creating it.
    
 	I’ve experimented with a number of methods over time and am seeing these real issues with each, so I found a middle ground that provides flexibility with a small redundancy, some optional values, without messing up the basic structure.  This is a relatively complex design which I’ve enjoyed working on, and I do care about what design works best, I just don’t think  the solution here is going to be cookie cutter. This is not my favorite design decision. I feel the need to neatly separate structs like I’ve done in my traditional OOP, but it's not possible. I’m learning that GO seems to enforce a different style which prefers redundancy and linearity over cyclicness and readability but I’m adapting to it. 

	I think I understand your concern about this particular aspect, but we can clarify  it in a pairing session after Boran is back. I’d love it if I can learn new things in a meet if I haven’t considered anything already. Otherwise, if this explanation addresses it, we can proceed with this. Of course any comments on other aspects of the diff are welcome.

pkg/cloudclient/cloudclient.go
ritmun: I wrote an explanation https://github.com/openshift/osd-network-verifier/pull/107#issuecomment-1171778922

cmd/egress/cmd.go
ritmun: I wrote an explanation https://github.com/openshift/osd-network-verifier/pull/107#issuecomment-1171778922

pkg/cloudclient/cloudclient.go
ritmun: Nope. Initially it was printing "options.Region" which will always print what you provide on cli. "getDefaultRegion" function will print the actual region being used. Whether in cli or in the env, or the default val set by code if neither exists.

fahlmant: Something is still not right here

```
$  git log | head -1
commit 97abcd959ceb4e285e78bf2e1a67e3b2737b0c3b

$ make build
go fmt ./...
go mod tidy
go build -mod=mod .


$ AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=bar ./osd-network-verifier egress --subnet-id subnet-00fdc1eac0be9449d  --region us-west-2
Using region: us-east-1
Using AWS secret key
```

ritmun:
[ reference PR https://github.com/ritmun/osd-network-verifier/pull/2  for demo only ]

ritmun:
This PR has become difficult to review due to addition of requested changes. I have separated the requested changes into their own PRs and I will be creating separate PRs from them, each referencing their own card. 

1. egress test for osdctl https://github.com/openshift/osdctl/pull/225 
2. profile flag in onv https://github.com/openshift/osd-network-verifier/pull/118 
3. PR for refactor will be created soon under this card https://issues.redhat.com/browse/OSD-12381

PR #108, opened 2022-06-09
Updated the userdata template to work properly with custom AMI

bdematte:
Few changes to allow proper work of osd-network-verifier with our AMI : 
- could not find a clean way to redirect all outputs to the console (found one requiring bash but could not get the generated usrcmd to not use sh) so staging the outputs to a file and sending it to /dev/console as last step
- removing the start of docker as it is failing on our custom AMI (and faillure is detected as 'not internet connection'). From quick investigations, seems docker is aliased to podman in our AMI (except for systemd entries)

### DIFF
--- a/pkg/helpers/config/userdata.yaml
+++ b/pkg/helpers/config/userdata.yaml
@@ -1,14 +1,10 @@
 #cloud-config
 repo_update: true
 package_update: true
-
-packages:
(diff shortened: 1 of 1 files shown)


fahlmant (approved):
/approve
/lgtm

PR #110, opened 2022-06-13
Osd 11484 change to generic error type

ritmun:
All error types are currently showing as EgressError.

Changed this to generic ONV error. Added unit tests for both error types.

To test locally:
1. make build
2. Disallow all inbound traffic to your subnet 
3.  ./osd-network-verifier egress --subnet-id your_subnet  --region your_region

should return:
 "network verifier error: internet connectivity problem: please ensure there's internet access in given vpc subnets"

Egress errors for specific endpoints should still show as egress URL error. 

4. To run tests `make test`
 
 Jira https://issues.redhat.com/browse/OSD-11484


pkg/output/output.go
```diff
@@ -30,13 +30,23 @@ func (o *Output) AddException(message error) {
...
 	}
 }
 
+// GetEgressFailures gets egress endpoint failures
+func (o *Output) GetEgressFailures() (failures []error) {
+	return o.failures
+}
+
+// GetExceptions gets generic errors
+func (o *Output) GetExceptions() (errors []error) {
```
boranx, line 46: why do we need this? I saw the usage only in _tests_
I can recommend using `Parse()` which returns all the output params individually
```go
_, exceptions, _ := cli.output.Parse()
```

pkg/output/output.go
```diff
@@ -30,13 +30,23 @@ func (o *Output) AddException(message error) {
...
-func (o *Output) SetFailures(failures []string) {
+// SetEgressFailures sets egress endpoint failures as a bulk update
+func (o *Output) SetEgressFailures(failures []string) {
 	for _, f := range failures {
 		o.failures = append(o.failures, handledErrors.NewEgressURLError(f))
 	}
 }
 
+// GetEgressFailures gets egress endpoint failures
+func (o *Output) GetEgressFailures() (failures []error) {
```
boranx, line 41: same as above: consider using `cli.output.Parse()`

pkg/cloudclient/aws/private_test.go
```diff
@@ -88,3 +90,102 @@ func TestValidateEgress(t *testing.T) {
...
+
+}
+
+//Test generic type errors
+func TestValidateGenericError(t *testing.T) {
+	vpcSubnetID, cloudImageID := "dummy-id", "dummy-id"
+	consoleOut := `[   48.062407] cloud-init[2472]: Cloud-init v. 19.3-44.amzn2 running 'modules:final' at Mon, 
+07 Feb 2022 12:30:22 +0000. Up 48.00 seconds.
+	[   48.077429] cloud-init[2472]: USERDATA BEGIN
+Could not do X.
```
boranx, line 151: (optional) in such tests, repeating the same scenario by just changing one field - instance console output/result correlation -  
we can consider making table-driven tests where we test output by adjusting _consoleOut_

pkg/errors/errors.go
```diff
@@ -22,13 +22,25 @@ func NewEgressURLError(failure string) error {
...
 }
 
-func (e *GenericError) Error() string { return e.e }
+func (e *GenericError) Error() string { return e.message }
 
-func NewGenericError(err error) error {
+func NewGenericError(message string) error {
 	return &GenericError{
-		e: fmt.Sprintf("generic(unhandled) error: %s ", err.Error()),
+		message: fmt.Sprintf("network verifier error: %s", message),
```
boranx, line 32: are we ok to start each one of the generic error with `network verifier error:`? Would we want to include _generic_ type in the message e.g `unhandled error` ?
To me `network verifier error` looks a bit odd 
1- because we know that error will be always thrown from `network verifier` in verifier execution context.
2- network verifier error sounds like it can be anything like egressurl error/generic error/ any other error that's defined

pkg/errors/errors.go
```diff
@@ -22,13 +22,25 @@ func NewEgressURLError(failure string) error {
 }
 
 type GenericError struct {
-	e string
+	message string
```
boranx, line 25: if we want to change `e` -> `message`, we should perform this to all EgressURLError, GenericError, or any other err we defined, in terms of standardization

boranx (changes requested):
Hey @ritmun
I've left some open questions, but in general the idea is ok

pkg/output/output.go
ritmun: I used it for asserting the unit test. same as above..

pkg/errors/errors.go
ritmun: _Would we want to include generic type in the message:_ Is "generic network verifier error" better?

_we know that error will be always thrown from network verifier in verifier execution context._  : Imo, including "network verifier" name in the error message gives a better context when it's integrated with other tools and produces output in the common stdout with other tools. 

_network verifier error sounds like it can be anything like egressurl error/generic error/ any other error that's defined_ I'm not sure I understand. EgressURL error will still show as EgressURL error.. it's only when we want generic error we can use this. Did I implement it in a place it should be more specific error?

pkg/errors/errors.go
```diff
@@ -22,13 +22,25 @@ func NewEgressURLError(failure string) error {
...
-func NewGenericError(err error) error {
+func NewGenericError(message string) error {
 	return &GenericError{
-		e: fmt.Sprintf("generic(unhandled) error: %s ", err.Error()),
+		message: fmt.Sprintf("network verifier error: %s", message),
+	}
+}
+
+type UnhandledError struct {
+	e string
```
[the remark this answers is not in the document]
ritmun: (Generic)  ONV error is an error we have handled, but we haven't designated a specific code to it (e.g. egress URL) 

Unhandled error should be the last catch for any error which isn't caught (by mistake or intention) and falls in an all-catching handler.  (conceptually equivalent of outermost try,catch)

We aren't using the unhandled error, but it was already there so I didn't want to remove it.

pkg/errors/errors.go
ritmun: Will do. I wasn't sure about that.

pkg/cloudclient/aws/private_test.go
```diff
@@ -88,3 +90,102 @@ func TestValidateEgress(t *testing.T) {
...
+	encodedconsoleOut := base64.StdEncoding.EncodeToString([]byte(consoleOut))
+	FakeEC2Cli := mocks.NewMockEC2Client(ctrl)
+	FakeEC2Cli.EXPECT().GetConsoleOutput(gomock.Any(), gomock.Any()).Times(1).Return(&ec2.GetConsoleOutputOutput{
+		Output: aws.String(encodedconsoleOut),
+	}, nil)
+	FakeEC2Cli.EXPECT().RunInstances(gomock.Any(), gomock.Any()).Times(1).Return(&ec2.RunInstancesOutput{
+		Instances: []types.Instance{{
+			InstanceId: aws.String(testID),
+		}},
+	}, nil)
```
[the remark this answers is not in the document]
ritmun: yes, we need that.

ritmun:
> Hi Ritu,
> 
> After executing the PR, it gives the below error.
> 
> printing out failures: printing out exceptions preventing onv from running:
> 
> * network verifier error: internet connectivity problem: please ensure there's internet access in given vpc subnets
>   printing out errors faced during the execution:
>   Failure!

Thanks for verifying!

pkg/cloudclient/aws/private_test.go
ritmun: done

pkg/errors/errors.go
boranx: Is "generic network verifier error" better? IMO, no :)) I think we can remove the generic one if we ok with network-verifier-error

boranx:
/hold
/lgtm

Feel free to `unhold` when you're ready :)

boranx (commented):
/approve

PR #111, opened 2022-06-15
Fixing script hanging in case of too many errors

bdematte:
- Updating the calculation of the error map length in order to account for domains being checked for multiple ports
- Increasing the Poll duration as faced some timeouts due to too low value
- Replacing the 'true' with an echo of an error message to ensure an Error is raised and detected in case the docker container didn't exit with 0

Tracking ticket : [OSD-11781](https://issues.redhat.com/browse/OSD-11781)


fahlmant (approved):
/approve
/lgtm

PR #112, opened 2022-06-17
Add configurable endpoint connection retry feature

abyrne55:
This PR adds retry behavior to the "network-validator" binary that gets baked into the Docker image, such that the binary will retry a failed endpoint connection attempt `maxRetries` times before declaring the endpoint unreachable. `maxRetries` is configurable via command line flag, and defaults to 3 attempts.

This addresses [OSD-12061](https://issues.redhat.com/browse/OSD-12061), where a lack of retry behavior has been found to cause false negatives (i.e., falsely claiming that an endpoint is blocked).

### DIFF
--- a/build/bin/network-validator.go
+++ b/build/bin/network-validator.go
@@ -21,6 +21,7 @@ import (
 
 var (
 	timeout        = flag.Duration("timeout", 1000*time.Millisecond, "Timeout for each dial request made")
+	maxRetries     = flag.Int("max-retries", 3, "Maximum connection attempts per endpoint")
 	configFilePath = flag.String("config", "config.yaml", "Path to configuration file")
 )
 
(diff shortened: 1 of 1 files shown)


boranx (commented):
/lgtm

thanks for that @abyrne55 :rocket:

bdematte:
It would be great to have the option on osd-network-verifier so that when using the tool, we can override the value directly.

abyrne55:
> It would be great to have the option on osd-network-verifier so that when using the tool, we can override the value directly.

Agreed, but I decided to limit the scope of this PR to just the validator binary for now, as adding a command line flag/API call argument is blocked until [ValidateEgress()](https://github.com/openshift/osd-network-verifier/blob/04e9ed47d945d8fc6bf6ff3f88bde1f8f73b186d/pkg/cloudclient/aws/aws.go#L45) has been refactored to use an options struct instead of individual parameters.

PR #113, opened 2022-06-21
Osd - 12063

lnguyen1401:
changes default timeout from 1s to 2s 
related jira card: https://issues.redhat.com/browse/OSD-12063


fahlmant:
@lnguyen1401 Was this branch based off the more current commit on the upstream `main` branch?

/hold

ritmun:
It'll do, but any way we can use a const for the duration instead of using the value? an example https://github.com/openshift/cloud-ingress-operator/blob/f3edb51b9e5ba9c6f797fb805681bf31f5ddb6b4/pkg/controller/apischeme/apischeme_controller.go#L36

fahlmant:
We want it configurable by flag, so a const wouldn't make sense here

ritmun:
> We want it configurable by flag, so a const wouldn't make sense here

 I missed that. Makes sense.

build/bin/network-validator.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	timeout        = flag.Duration("timeout", 1000*time.Millisecond, "Timeout for each dial request made")
+	timeout        = flag.Duration("timeout", 2000*time.Millisecond, "Timeout for each dial request made")
```
ritmun, line 23: Looks like this value matches the default for the var. Slight chance of the two falling out of sync. Any chance this could be a central config value? And use that const as default in cmd as well as this?

boranx: yes, This parameter also exists in https://github.com/openshift/osd-network-verifier/blob/main/pkg/cloudclient/cloudclient.go#L28 and allows us to pass it to `network-validator.go`'s dockerised form as to be run by `userdata` in bootstrap step of the ec2 instance.

I think that it makes sense to set it to a time greater than 1

boranx (commented):
/lgtm
/hold

The change is pretty straightforward, so should be good to go.
Thanks for the PR @lnguyen1401

abyrne55 (approved):
/lgtm

Tested on my dev account; works as intended. Nice work!

build/bin/network-validator.go
ritmun: Ok

PR #115, opened 2022-07-04
OSD-11931 - Fixing osd-network-verifier to work properly with our golden AMI as well

bdematte:
Changes have been done in order to work properly with our golden AMI and improve robustness : 
- Silencing the start of docker service (which was triggering false positive due to our handling of error, should be improved in OSD-12345)
- Managing packages in the `runcmd` section to be able to make it non-blocking (to take advantage of the pre-install docker in our golden AMI)

Reference : [OSD-11931](https://issues.redhat.com/browse/OSD-11931)

### DIFF
--- a/pkg/helpers/config/userdata.yaml
+++ b/pkg/helpers/config/userdata.yaml
@@ -1,13 +1,13 @@
 #cloud-config
-repo_update: true
-package_update: true
-packages:
-  - docker 
 runcmd:
-  - sudo service docker start
+  # taking care manually of the packages to be able to make it non-blocking
+  - sudo yum -y update || echo "didn't update package"
+  - sudo yum -y install docker || echo "docker not installed by yum"
+  # Redirecting the output because of failed start with Golden AMI (should be improved in OSD-12345)
+  - sudo service docker start 2>1 > /dev/null || echo "docker not started by systemctl"
   - echo "${USERDATA_BEGIN}" >> /var/log/userdata-output
   - sudo docker pull ${VALIDATOR_IMAGE}
-  # Use `|| true` to ignore failure exit codes, we want the script to continue either way
+  # Printing a message with 'Failed' in order to ensure the is detected as a failure by osd-network-verifier (OSD-12345 to improve this)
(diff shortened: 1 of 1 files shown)


pkg/helpers/config/userdata.yaml
```diff
@@ -1,13 +1,13 @@
 #cloud-config
-repo_update: true
-package_update: true
-packages:
-  - docker 
 runcmd:
-  - sudo service docker start
+  # taking care manually of the packages to be able to make it non-blocking
```
ritmun, line 3: Would it be ok to make updates non-blocking if we want to enforce an update e.g. if there's a security update included? Can we switch blocking failed updates using an input flag?

bdematte: We can card it if we feel it is needed, but I didn't as I don't believe there is a strong usecase : 
- the default AMI will be used by OCM without updates, so we need to be fine with using this AMI on customer account
- when using the cli, we already have the ability to specify the ami to use if we face a specific scenario (and golden-ami is installing up-to-date packages).

As a side note, I still need to create a card for the maintenance of the AMI with regard to security patchs and to get rid of the explicit AMIs in the code which is the longer term solution for the security patching. But I am not sure this should be blocking here.

pkg/helpers/config/userdata.yaml
```diff
@@ -1,13 +1,13 @@
...
+  - sudo yum -y update || echo "didn't update package"
```
boranx, line 4: do we need `update` and `install docker` given it'll be already done on AMI?

pkg/helpers/config/userdata.yaml
```diff
@@ -1,13 +1,13 @@
...
+  - sudo yum -y install docker || echo "docker not installed by yum"
+  # Redirecting the output because of failed start with Golden AMI (should be improved in OSD-12345)
+  - sudo service docker start 2>1 > /dev/null || echo "docker not started by systemctl"
```
boranx, line 7: we should embed this to AMI. Also we can add `sudo docker pull ${VALIDATOR_IMAGE}` part to AMI to get a faster validation check

pkg/helpers/config/userdata.yaml
bng0y: because we give the option to provide different AMIs, we likely still have to do this. For e.g when Cu provides an ami without docker installed.

bng0y: maybe we could skip it, when using our AMIs

PR #116, opened 2022-07-05
added // #nosec for InsecureSkipVerify

lnguyen1401:
added #nosec for InsecureSkipVerify to skip this check for the linter as a result of gosec :

test "/home/linnguye/golangci.yml" = "" || test ! -e "/home/linnguye/golangci.yml" || GOLANGCI_LINT_CACHE="/tmp/golangci-cache" golangci-lint run -c "/home/linnguye/golangci.yml" ./...
build/bin/network-validator.go:119:33: G402: TLS InsecureSkipVerify set true. (gosec)
			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
			                        
Justification: Low chance of MITM, as the instance is short-lived
related Jira card: https://issues.redhat.com/browse/OSD-11465

### DIFF
--- a/build/bin/network-validator.go
+++ b/build/bin/network-validator.go
@@ -115,6 +115,7 @@ func ValidateReachability(host string, port int, tlsDisabled bool) error {
 		Timeout: *timeout,
 	}
 
+	// #nosec G402 -- Low chance of MITM, as the instance is short-lived, see OHSS-11465
 	if tlsDisabled {
 		httpClient.Transport = &http.Transport{
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},


fahlmant (commented):
@lnguyen1401 Did you run the verifier with this change to make sure there are no issues? Also looks like `go mod tidy` needs to be ran.

build/bin/network-validator.go
```diff
@@ -115,6 +113,7 @@ func ValidateReachability(host string, port int, tlsDisabled bool) error {
 		Timeout: *timeout,
 	}
 
+	// #nosec G402 -- Low chance of MITM, as the instance is short-lived, see OHSS-11465
```
fahlmant, line 116: Did you mean to add this in this PR?

lnguyen1401: yeah, I think i need to add that to skip the gosec as Jason commented on the card:
`[Linh Nguyen](https://issues.redhat.com/secure/ViewProfile.jspa?name=linnguye.openshift), I agree with your justification about disabling that check. My preference for disabling any checks is that the specific lines or blocks have the specific gosec check disabled — with the justification inline or cited — rather than disabling the check completely in the gosec options.`

lnguyen1401:
Ran the verifier with the change. all good!

build/bin/network-validator.go
fahlmant: Can we please separate these into two different PRs?

fahlmant (approved):
/approve
/lgtm

PR #124, opened 2022-07-27
Add more explicit error messages for generic errors and docker errors

fahlmant:
The verifier currently gives an ambiguous error when a "generic" error happens. This PR adds two things to help clarify how to handle non-test failures:

1. Adds a check to see if the problem is specifically related to docker, and a specific message to indicate what went wrong. Error message is added to the print out.
2. Refines the generic error message to indicate that more investigation is needed as the egress test was not able to run fully.


sam-nguyen7 (commented):
/lgtm

Much needed checks to clarify how we handle failures. Thanks!

clcollins:
Late to the party LGTM - thanks for the extra clarification!  This will help a lot, thank you.

PR #135, opened 2022-09-20
OSD-13241 Add debug logs to output struct

mjlshen:
The output now looks something like this when using the `--debug` flag. It looks the same as before without the debug flag.
```
Summary:
printing out debug logs from the execution:
 -  Gathering description of instance type t3.micro from EC2
 -  Using configured timeout of 2s for each egress request
 -  base64-encoded generated userdata script:
---
${b64UserData}
---
 -  Using AMI: ami-09d6a8053437e16bf
 -  Waiting for EC2 instance i-0cbce467076c2637d to be running
 -  Describing state of EC2 instance i-0cbce467076c2637d
 -  Instance i-0cbce467076c2637d has no status yet
 -  Describing state of EC2 instance i-0cbce467076c2637d
 -  Scraping console output and waiting for user data script to complete...
 -  EC2 console consoleOutput contains data, but end of userdata script not seen, continuing to wait...
 -  EC2 console consoleOutput contains data, but end of userdata script not seen, continuing to wait...
 -  base64-encoded console logs:
---
${b64ConsoleLogs}
---
All tests pass!
Success
```

### DIFF
--- a/cmd/dns/cmd.go
+++ b/cmd/dns/cmd.go
@@ -67,7 +67,7 @@ func NewCmdValidateDns() *cobra.Command {
 			}
 
 			out := cli.VerifyDns(ctx, config.vpcID)
-			out.Summary()
+			out.Summary(config.debug)
 			if !out.IsSuccessful() {
 				logger.Error(ctx, "Failure!")
 				os.Exit(1)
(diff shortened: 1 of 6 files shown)


boranx (commented):
/lgtm
/approve

just a small note, it looked a bit more verbose after I run the new build against my subnet, so in the future, it'd be nice to have debug levels e.g --debug 1 --debug 2 or some other mechanisms
but the idea is to hide  `base64-encoded console logs` unless it's a very verbose mode

PR #139, opened 2022-09-22
Bugfix - correct name tag capitalization

mjlshen:
On-call ran into this issue: 
`fix for network verifier error: error performing ec2:CreateTags: For 'Name', use the capitalization specified by the tag policy.`

This is due to an AWS tag policy specified by the customer (https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_example-tag-policies.html), but we should use the correct capitalization anyway.

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -16,7 +16,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


fahlmant (approved):
/approve

PR #145, opened 2022-10-06
Adding Security group

anispate:
OSD-12357


pkg/verifier/aws/entry_point.go
```diff
@@ -76,15 +93,6 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		return a.Output.AddError(err) // fatal
 	}
 
-	if instanceReadyErr := a.AwsClient.WaitForEC2InstanceCompletion(vei.Ctx, instanceID); instanceReadyErr != nil {
```
fahlmant, line 79: Was this removal on purpose? It needs to remain in the code

pkg/verifier/aws/aws_verifier.go
```diff
@@ -179,6 +179,12 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
 		return "", handledErrors.NewGenericError(err)
 	}
 
+	// Wait up to 5 minutes for the instance to be running
+	waiter := ec2.NewInstanceRunningWaiter(a.AwsClient)
```
fahlmant, line 183: What led to this being added?

pkg/clients/aws/aws.go
```diff
@@ -85,63 +92,27 @@ func (c *Client) GetConsoleOutput(ctx context.Context, input *ec2.GetConsoleOutp
...
+	return c.ec2Client.CreateSecurityGroup(ctx, params, optFns...)
+}
 
-	return &result.InstanceStatuses[0].InstanceState.Name, nil
+func (c *Client) DescribeSubnets(ctx context.Context, params *ec2.DescribeSubnetsInput, optFns ...func(*ec2.Options)) (*ec2.DescribeSubnetsOutput, error) {
+	return c.ec2Client.DescribeSubnets(ctx, params, optFns...)
 }
 
-// waitForEC2InstanceCompletion checks every 15s for up to 2 minutes for an instance to be in the running state
-func (c *Client) WaitForEC2InstanceCompletion(ctx context.Context, instanceID string) error {
```
fahlmant, line 121: This should be used by the egress test

pkg/verifier/aws/aws_verifier.go
```diff
@@ -318,3 +324,120 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+		Description: awsTools.String("osd-network-verifier security group"),
+	}
+	output, err := a.AwsClient.CreateSecurityGroup(ctx, input)
+	if err != nil {
+		return nil, err
+	}
+
+	a.Logger.Info(ctx, "Created security group with ID: %s", *output.GroupId)
+
+	input_rules := &ec2.AuthorizeSecurityGroupEgressInput{
```
fahlmant, line 342: This setting is likely too permissive. If possible, this should match the security group that the OpenShift installer creates

fahlmant (changes requested):

pkg/verifier/aws/entry_point.go
anispate: We are replacing this with the NewInstanceRunningWaiter which waits for the instance to be running for upto 5 mins.

pkg/verifier/aws/aws_verifier.go
mjlshen: Ah it's a native `aws-sdk-go-v2` way to "wait" for things to happen https://aws.github.io/aws-sdk-go-v2/docs/making-requests/#using-waiters which let us deprecate the custom logic we had previously in `WaitForEC2InstanceCompletion` and `describeEC2Instances`

mjlshen: The cost being a permissions switch from `ec2:DescribeInstanceStatus` to `ec2:DescribeInstances`

fahlmant: @mjlshen Makes sense. However, I think this should be done in a separate PR.

fahlmant:
@anispate Can you please add a description of what changed and why to the description?

As far as the changes, I think it would be cleaner to move the ec2 logic changes to a separate PR.

pkg/verifier/aws/aws_verifier.go
mjlshen: Ah this is a least-privilege one defined by https://issues.redhat.com/browse/OSD-12356, however we definitely need to tag this security group with `"red-hat-managed: true"`

* No ingress allowed
* Allowed egress to 0.0.0.0/0 for ports 80/443/9997 since we can't restrict egress by domain name

bdematte:
We need some parameter on the client to enable it : 
- when using the cli, it will be used only with Support role which doesn't have the 'create SG' priviledge (and we don't want to add to to support role). Especially, for debugging an existing cluster, it is not relevant to create a security group (we should be using existing ones, we have a parameter for this, and a card to automatically detect the worker SG to be implemented later)
- the case to create the SG is basically only when triggered by OCM.

Probably there should be a 'create SG' parameter in the config which is set to true by default (to create a SG without changes on OCM side) but which is overridden to false by the cli by default (not to create any SG when triggered from the client).

pkg/verifier/aws/aws_verifier.go
```diff
@@ -318,3 +324,120 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+	}
+
+	if _, err := a.AwsClient.RevokeSecurityGroupEgress(ctx, revoke_default_egress); err != nil {
+		return nil, err
+	}
+
+	return output, nil
+}
+
+// DeleteSecurityGroup deletes a security group with the specified ID
```
rafael-azevedo, line 405: We can put this right in the main logic. Its just setting up inputs and calling the client . Since we abstract it away in the awclient package the "DeleteSecruityGroups" function we don't need this abstraction in here for testing either.  Also is the function used ?

pkg/verifier/aws/entry_point.go
rafael-azevedo: @anispate is this replacement going to be part of this PR ?

pkg/clients/aws/aws.go
```diff
@@ -150,5 +121,11 @@ func (c *Client) TerminateEC2Instance(ctx context.Context, instanceID string) er
 		return handledErrors.NewGenericError(err)
 	}
 
+	// Wait up to 5 minutes for the instance to be terminated
+	waiter := ec2.NewInstanceTerminatedWaiter(c)
+	if err := waiter.Wait(ctx, &ec2.DescribeInstancesInput{InstanceIds: []string{instanceID}}, 5*time.Minute); err != nil {
```
rafael-azevedo, line 126: How did we arrive at the 5 minute wait time ? Is this what the previous behavior did ?

mjlshen (changes requested):

pkg/verifier/aws/aws_verifier.go
mjlshen: It's used ~line 105 of pkg/verifier/aws/entry_point.go

pkg/verifier/aws/aws_verifier.go
```diff
@@ -129,8 +129,10 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
 	}
 
 	eniSpecification := ec2Types.InstanceNetworkInterfaceSpecification{
-		DeviceIndex: awsTools.Int32(0),
-		SubnetId:    awsTools.String(input.SubnetID),
+		// Required for when the EC2 instance is in a public subnet
+		AssociatePublicIpAddress: awsTools.Bool(true),
```
fahlmant, line 133: This needs to be removed

pkg/verifier/aws/aws_verifier.go
```diff
@@ -325,3 +327,120 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
 	a.Output.AddDebugLogs(log)
 	a.Logger.Debug(context.TODO(), log)
 }
+
+// CreateSecurityGroup creates a security group with the specified name and cluster tag key in a specified VPC
+func (a *AwsVerifier) CreateSecurityGroup(ctx context.Context, name, vpcId string) (*ec2.CreateSecurityGroupOutput, error) {
+	input := &ec2.CreateSecurityGroupInput{
+		GroupName:   awsTools.String(name),
+		VpcId:       &vpcId,
+		Description: awsTools.String("osd-network-verifier security group"),
```
fahlmant, line 336: Please add `TagSpecification` and fill with the tags provided

fahlmant (changes requested):

pkg/verifier/aws/aws_verifier.go
mjlshen: Done in #149

pkg/verifier/aws/entry_point.go
mjlshen: Done in #149

pkg/verifier/aws/aws_verifier.go
anispate: Part of the latest commit

pkg/verifier/aws/aws_verifier.go
rafael-azevedo: @mjlshen @anispate I think we should still remove this func unnecessary abstraction just call it directly in the entry_point with the client.

pkg/verifier/aws/aws_verifier.go
```diff
@@ -325,3 +325,134 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+// CreateSecurityGroup creates a security group with the specified name and cluster tag key in a specified VPC
+func (a *AwsVerifier) CreateSecurityGroup(ctx context.Context, tags map[string]string, name, vpcId string) (*ec2.CreateSecurityGroupOutput, error) {
+	input := &ec2.CreateSecurityGroupInput{
+		GroupName:   awsTools.String(name),
+		VpcId:       &vpcId,
+		Description: awsTools.String("osd-network-verifier security group"),
+	}
+	output, err := a.AwsClient.CreateSecurityGroup(ctx, input)
+	if err != nil {
+		return nil, err
```
rafael-azevedo, line 338: all these nils should be &ec2.CreateSecurityGroupOutput{} to avoid possibility of nil pointer.

anispate: Its updated as part of the last commit

pkg/verifier/aws/aws_verifier.go
```diff
@@ -325,3 +325,134 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+	// Wait up to 1 minutes for the security group to exist
+	waiter := ec2.NewSecurityGroupExistsWaiter(a.AwsClient)
+	if err := waiter.Wait(ctx, &ec2.DescribeSecurityGroupsInput{GroupIds: []string{*output.GroupId}}, 1*time.Minute); err != nil {
+		if _, err := a.DeleteSecurityGroup(ctx, *output.GroupId); err != nil {
+			return nil, handledErrors.NewGenericError(err)
+		}
+		return nil, fmt.Errorf("deleted %s after timing out waiting for security group to exist", *output.GroupId)
+	}
+
+	if err := a.createTags(tags, *output.GroupId); err != nil {
```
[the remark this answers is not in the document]
anispate: Completed

rafael-azevedo (approved):
/lgtm

pkg/verifier/aws/aws_verifier.go
```diff
@@ -325,3 +325,133 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+	if err != nil {
+		return &ec2.CreateSecurityGroupOutput{}, err
+	}
+
+	a.Logger.Info(ctx, "Created security group with ID: %s", *output.GroupId)
+
+	// Wait up to 1 minutes for the security group to exist
+	waiter := ec2.NewSecurityGroupExistsWaiter(a.AwsClient)
+	if err := waiter.Wait(ctx, &ec2.DescribeSecurityGroupsInput{GroupIds: []string{*output.GroupId}}, 1*time.Minute); err != nil {
+		if _, err := a.AwsClient.DeleteSecurityGroup(ctx, &ec2.DeleteSecurityGroupInput{
```
fahlmant, line 346: Will this work if the SG never gets created? I think it would just error saying "SG doesn't exist" and cause the egress test to fail.

pkg/verifier/aws/entry_point.go
```diff
@@ -61,6 +61,23 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
...
 	}
 
+	cleanupSecurityGroup := false
+	if vei.AWS.SecurityGroupId == "" {
+		vpcId, err := a.GetVpcIdFromSubnetId(vei.Ctx, vei.SubnetID)
+		if err != nil {
+			return a.Output.AddError(err)
+		}
+
+		createSecurityGroupOutput, err := a.CreateSecurityGroup(vei.Ctx, vei.Tags, "ONV_security_group", vpcId)
```
fahlmant, line 71: Can the name passed in here please match the name of the EC2 instance as well?

docs/aws/aws.md
```diff
@@ -60,7 +60,13 @@ Ensure that the AWS credentials being used have the following permissions. (This
         "ec2:DescribeInstanceTypes",
         "ec2:GetConsoleOutput",
         "ec2:TerminateInstances",
-        "ec2:DescribeVpcAttribute"
+        "ec2:DescribeVpcAttribute",
+        "ec2:CreateSecurityGroup",
```
bdematte, line 64: As SREs running manually the checks, we don't have (and don't want/won't have Create/Delete SG permissions).
This should be clarified in the docs. 
As a SRE, the expected flow is to specify the security group on the command line (as the default SG may not be suitable and does not correspond to what is used by OCP anyway). As discussed, the verifier shouldn't run in case no security group is provided and nicely error. 
For the node, we have a card already tracking the implementation of smart retrieve of the SG to be used to avoid the user to have to connect to AWS and retrieve it by him/herself (tracked in OSD-12250).

bdematte (changes requested):

pkg/verifier/aws/aws_verifier.go
mjlshen: @anispate will test this by 

* On line 343 (above this comment), we can try this by adding something like `time.Sleep(1 * time.Minute)`
* Then test ONV, let it run, it’ll create a security group and then sleep for 1 minute
* During this time, have AWS ready and delete the security group in the UI
* See what error ONV returns when it wakes up from sleep and tries to describe/delete the security group. Improve the error message if needed

anispate: printing out errors faced during the execution:
 -  network verifier error: error performing ec2:DeleteSecurityGroup: The security group 'sg-0631479082ecd80c6' does not exist
Failure!

Hi Taylor, throws this error.

fahlmant: Yeah, this is what I expected but I think its a bad message to send, since the real error is the SG was unable to create.

anispate: So, the way. i tested it is by what Michael defined above....

SG was created then i went to AWS UI and deleted the SG from there which resulted in the SG doesn't exist. 

But, initially, the SG was created.

fahlmant: Right, but you were emulating the case where the SG wasn't actually created in the first place.

mjlshen: This is a bit different I think, it's an edge case where AWS receives and validates a request to create security group and then fails to do so within one minute. We could prepend any errors in this block with "expected security group %s to exist, but did not find it within 1 minute" or something? Or maybe just add debug logs when we create, wait, and delete security groups?

In the more common case where AWS receives a request to create a security group and ONV is missing permissions or it's a malformed request, a relevant error would be returned about failing to create the security group

docs/aws/aws.md
anispate: Added it as part of the latest PR

pkg/verifier/aws/entry_point.go
anispate: Added it as part of the latest PR

pkg/verifier/aws/aws_verifier.go
anispate: It was taken care of as part of the latest PR.

pkg/verifier/aws/aws_verifier.go
```diff
@@ -325,3 +325,135 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+	}
+
+	// What if we get an empty vpc-id for a returned subnet
+	if len(output.Subnets) == 0 {
+		return "", fmt.Errorf("no subnets returned for subnet id: %s", vpcSubnetID)
+	}
+
+	// What if the Subnets array has 0 entries
+	vpcId := *output.Subnets[0].VpcId
+	if len(vpcId) == 0 {
```
fahlmant, line 454:
```suggestion
	if vpcId == "" {
```

cmd/egress/cmd.go
```diff
@@ -217,5 +217,10 @@ are set correctly before execution.
 		os.Exit(1)
 	}
 
+	if err := validateEgressCmd.MarkFlagRequired("security-group-id"); err != nil {
```
boranx, line 220: if this will be transformed into a required field, we can mention it on the cmd usage doc, too
```shell
▶./osd-network-verifier egress --subnet-id test
Error: required flag(s) "security-group-id" not set
Usage:
  osd-network-verifier egress [flags]

Examples:
For AWS, ensure your credential environment vars
AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY (also AWS_SESSION_TOKEN for STS credentials)
are set correctly before execution.

# Verify that essential OpenShift domains are reachable from a given SUBNET_ID
./osd-network-verifier egress --subnet-id $(SUBNET_ID)
```

and note that the error here is for reporting the registration error, not when the flag is not provided
so IMO the following would be more graceful (+ to prevent the usage of os.exit around)
```
	if err := validateEgressCmd.MarkFlagRequired("security-group-id"); err != nil {
	}
```

anispate: So, removed os.Exit(1)?

pkg/verifier/aws/entry_point.go
```diff
@@ -84,6 +101,14 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		a.Output.AddError(err)
 	}
 
+	if cleanupSecurityGroup {
+		if _, err := a.AwsClient.DeleteSecurityGroup(vei.Ctx, &ec2.DeleteSecurityGroupInput{
+			GroupId: awsTools.String(vei.AWS.SecurityGroupId),
+		}); err != nil {
+			a.Output.AddError(err)
+		}
```
rafael-azevedo, line 109:
```suggestion
			_, err := a.AwsClient.DeleteSecurityGroup(vei.Ctx, &ec2.DeleteSecurityGroupInput{GroupId: awsTools.String(vei.AWS.SecurityGroupId)})
		if err != nil {
			a.Output.AddError(err)
		}
```

rafael-azevedo: I think it would read cleaner here with the if before the err statement instead of the before the return parameters.

pkg/verifier/aws/aws_verifier.go
```diff
@@ -325,3 +325,129 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+	if err != nil {
+		return &ec2.CreateSecurityGroupOutput{}, err
+	}
+
+	a.writeDebugLogs(fmt.Sprintf("Waiting for the Security Group to exist: %s", *output.GroupId))
+	// Wait up to 1 minutes for the security group to exist
+	waiter := ec2.NewSecurityGroupExistsWaiter(a.AwsClient)
+	if err := waiter.Wait(ctx, &ec2.DescribeSecurityGroupsInput{GroupIds: []string{*output.GroupId}}, 1*time.Minute); err != nil {
+		a.writeDebugLogs(fmt.Sprintf("Error waiting for the security group to exist: %s, attempting to delete the Security Group", *output.GroupId))
+		if _, err := a.AwsClient.DeleteSecurityGroup(ctx, &ec2.DeleteSecurityGroupInput{
```
rafael-azevedo, line 347: Same comment here about the if before the return paramertes

pkg/verifier/aws/aws_verifier.go
```diff
@@ -325,3 +325,129 @@ func (a *AwsVerifier) writeDebugLogs(log string) {
...
+		}); err != nil {
+			return &ec2.CreateSecurityGroupOutput{}, handledErrors.NewGenericError(err)
+		}
+		return &ec2.CreateSecurityGroupOutput{}, fmt.Errorf("deleted %s after timing out waiting for security group to exist", *output.GroupId)
+	}
+
+	a.Logger.Info(ctx, "Created security group with ID: %s", *output.GroupId)
+	if err := a.createTags(tags, *output.GroupId); err != nil {
+		// Unable to tag the instance
+		if _, err := a.AwsClient.DeleteSecurityGroup(ctx, &ec2.DeleteSecurityGroupInput{
```
rafael-azevedo, line 358: I think it would read cleaner here with the if before the err statement instead of the before the return parameters same as the others

rafael-azevedo (approved):
/lgtm

fahlmant (approved):
/lgtm
/approve
/hold cancel

PR #146, opened 2022-10-11
Sometimes the firewall can take more than 10 minutes to delete

mjlshen:
While running this integration test in all regions for [OSD-13240](https://issues.redhat.com//browse/OSD-13240) to validate the custom AMI, sometimes the firewall took more than 10 minutes to delete, so bumping to 20 minutes.


boranx (commented):
/lgtm
/approve

thanks for the fix!

PR #155, opened 2022-11-09
OSD-13207 adding grafana dashboard for osd-network-verifier

luis-falcon:
Feature

Following conversations with app-sre team it was suggested to create this file in order to set here the dashboards for the osd-network-verifier.  Thread https://coreos.slack.com/archives/CCRND57FW/p1667831061874529

Related issue OSD=13207

The PR contains a generated configmap that holds the promQL code for dashboards and queries that are run to create the dashboard in

### DIFF
--- a/dashboards/grafana-dashboard-network-verifier.configmap.yaml
+++ b/dashboards/grafana-dashboard-network-verifier.configmap.yaml
@@ -1,0 +1,653 @@
+apiVersion: v1
+data:
+  verifier-dashboard.json: |-
+    {
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
+            "datasource": {
+              "type": "datasource",
+              "uid": "grafana"
+            },
+            "enable": true,
+            "hide": true,
+            "iconColor": "rgba(0, 211, 255, 1)",
+            "name": "Annotations & Alerts",
+            "target": {
+              "limit": 100,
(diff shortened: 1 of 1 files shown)


dashboards/grafana-dashboard-network-verifier.configmap.yaml
```diff
@@ -0,0 +1,653 @@
...
+          "spaceLength": 10,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "gH4yqWBnz"
+              },
+              "expr": "(sum(network_verifier_failures{}) / sum(network_verifier_runs)) * 100",
```
boranx, line 147: just for knowledge: how are those metrics being populated?
are those from ocm backend? if so why do we integrate it into verifier's repo instead of the consumer service repo?

luis-falcon: those we built with the grafana playground. following the documentation for app-interface here : https://gitlab.cee.redhat.com/service/app-interface#add-a-grafana-dashboard 

Also the reason why we add it here is on the recommendation of the app-sre team the conversation is on the slack thread I put on the  description.

boranx (commented):
/lgtm
/approve

PR #156, opened 2022-11-10
chore: remove unnecessary parts from verifier, update docs

boranx:

This is a clean-up that removes unimplemented parts from the verifier.
Currently, looking at the codebase, some functions are left from the times when onv was just a poc, so I'd like to propose removing those. Those can be re-added when fully implemented
Also, it updates the doc to reflect verifier latest shape

integration tests were passed on local

## Checklist

- [x] I have made corresponding changes to the documentation

## Reviewer's Checklist


mjlshen (approved):
/lgtm

Thanks for this cleanup!

PR #157, opened 2022-11-17
Enhancement: This closes OSD-12772. Added flag to pass vpc-name for the gcp workflow

luis-falcon:
- Enhancement/Feature

This closes OSD-12772. Added flag to pass vpc-name for the gcp workflow

I added a flag to the egress command to pass the optional "vpc-name" flag explaining this will be part of the GCP workflow (checked only when the --gcp=true flag is set). In addition I left the other option of also having the env var GCP-VPC-NAME flag set so it can be one or the other.

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -41,6 +41,7 @@ type egressConfig struct {
 	noTls           bool
 	gcp             bool
 	awsProfile      string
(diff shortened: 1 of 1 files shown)


mjlshen (approved):
/lgtm

rafael-azevedo (approved):
/lgtm

PR #159, opened 2022-11-21
fix: pass https proxy if provided even when there's no ca

boranx:
Fix: https://issues.redhat.com/browse/OSD-13937

Enables the `https` if the `https proxy` is passed as transparent proxy (when there's no ca)

Testing: it can be tested against the cluster in the ticket
```shell
./osd-network-verifier egress --subnet-id subnet-xx --security-group-id sg-xxx --http-proxy "http://lookup-from-cluster:3128" --https-proxy "http://lookup-from-cluster:3128" --region eu-west-2
```


mjlshen (approved):
/lgtm

PR #160, opened 2022-11-23
fix: mark the execution as succeed if validator returns success

boranx:
This PR is remediation until https://issues.redhat.com/browse/OSD-12345 takes place

Currently, failure detection logic doesn't care whether the validator returned success or not
This fixes that to be checked in the beginning, else the failures will be searched and tried to be shown

Example execution where the docker is not installed properly (but embedded in the ami so validator still run)

```shell
./osd-network-verifier egress --subnet-id  subnet-005956b6508de4e84 --security-group-id sg-0891d931c405ac97b --region eu-west-1
Using region: eu-west-1
Created instance with ID: i-04385a55ff3494b85
Summary:
All tests pass!
Success
```


pkg/verifier/aws/aws_verifier.go
```diff
@@ -247,6 +248,12 @@ func (a *AwsVerifier) findUnreachableEndpoints(ctx context.Context, instanceID s
 				return false, nil
 			}
 
+			// Check if the result is success
```
mjlshen, line 251: This makes good sense, but do we still want to return `return true, nil` on line 285 with this new addition?

My understanding is:
- If `Success!`, then no error
- If it matches some other error, print accordingly
- If it gets to the end here - should we return an error now that all successes should be caught in the first branch? Previously if it got the end that meant that there was no error, therefore success.

mjlshen (approved):
/hold
/lgtm

I consider my comment a nit. If you feel like my suggested change isn't necessary feel free to remove my hold - I think this definitely addresses the intent of the improvement.

pkg/verifier/aws/aws_verifier.go
boranx: +1 if there's no `success` or `Unable to reach` either, it should return a generic err

boranx: hey @mjlshen I've covered this step w/ unit tests. On the line 278 we still want to `return true, nil` in order to halt the loop because we have `userdata end` but at line 285, I've changed the message to reflect an error because there's no occurrence in any of the regex patterns

boranx: hey @mjlshen I've added some unit tests to cover some scenarios,
by `return true, nil`, do you mean on line 278 or 285? (because on 285 I see we return false, nil probably to wait for consoleOutput.Output to be populated) I think due to poller function design, it returns at the end true and no err bc we don't know what we have in the output object (regex match)

boranx: on 278, I think it's necessary to use `return true, nil` to halt the loop as we have userdata end & regex matches stored in the output

mjlshen: Yeah, sorry you're right - line 278 is the right `return true, nil` I was wondering about.

I was wondering what you thought about something like `return true, errors.New("userdata script completed, but did not succeed due to an unknown error")` because at that point it would have seen userdata end, but not seen a "Success!", and also none of the existing regex checks for errors.

boranx: hmm I'm a bit confused, wouldn't there be also cases in which we hit the following?
```go
a.Output.SetEgressFailures(reUnreachableErrors.FindAllString(string(scriptOutput), -1))
```

I mean there could be just egress failures that were caught so on the next line we'd want a `return true, nil` ?

mjlshen: Yeah, that makes sense - thanks for explaining that!

mjlshen (approved):
/lgtm
/unhold

PR #161, opened 2022-12-01
Add additional helper function for setting up AwsVerifier

mjlshen:
For [OSD-14143](https://issues.redhat.com//browse/OSD-14143):

Adding a new helper function so consumers can just pass in their own AWS client/logger and not depend on network-verifier's opinionated workflow for assembling them. This will be useful in cases like osdctl and other consumers

### DIFF
--- a/pkg/verifier/aws/aws_verifier.go
+++ b/pkg/verifier/aws/aws_verifier.go
@@ -61,8 +61,23 @@ type AwsVerifier struct {
 	Output    output.Output
 }
 
+// GetAMIForRegion returns the default AMI given a region.
+// This is unused within this codebase, but exported so that consumers can access the values of defaultAmi
 func GetAMIForRegion(region string) string {
 	return defaultAmi[region]
+}
+
+// NewAwsVerifierFromConfig assembles an AwsVerifier given an aws-sdk-go-v2 config and an ocm logger
+func NewAwsVerifierFromConfig(cfg awsTools.Config, logger ocmlog.Logger) (*AwsVerifier, error) {
+	awsClient, err := aws.NewClientFromConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return &AwsVerifier{
+		AwsClient: awsClient,
+		Logger:    logger,
+	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


mjlshen:
This doesn't seem sufficient because it expects a network-verifier specific `*aws.Client`, this isn't aws-sdk-go's

fahlmant:
/hold
This needs further discussion

pkg/verifier/aws/aws_verifier.go
```diff
@@ -61,8 +61,17 @@ type AwsVerifier struct {
 	Output    output.Output
 }
 
-func GetAMIForRegion(region string) string {
```
fahlmant, line 64: Note from call today: OCM is using this and can't be removed.

mjlshen: Added back, with a comment explaining its purpose so it won't get deleted in the future

abyrne55 (changes requested):
Created a [test script](https://gist.github.com/abyrne55/8ef6d5f968e574557a2f8e17911c35c9) for this, and functionally it worked as expected. Nice work!

My only concern is regarding naming: `NewClientFromConfig()` makes sense, as it's creating an AWS client from an AWS config. `NewAwsVerifierFromClient()`, OTOH, leads me to assume that this function takes an AWS client (i.e., the output of `NewClientFromConfig()`) in order to return an AwsVerifier object. `Client` having two different meanings here (as `NewAwsVerifierFromClient()` takes in an AWS config object, _not_ an AWS client) is pretty confusing, so I suggest renaming this function to `NewAwsVerifierFromConfig`.

mjlshen:
Yeah, great point about the naming of the functions - I've renamed it!

boranx (commented):
/lgtm

thanks for the ability, is this still on hold? what was the outcome of the sync meeting?

boranx:
/lgtm

the bot didn't catch this as I previously made a mistake, so one more :)

mjlshen:
Roughly the outcome of the sync was that the use-case makes sense/doesn't interfere with the integration with OCM, but that it may be refactored away in the future. cc @rafael-azevedo

abyrne55 (approved):

rafael-azevedo (approved):

PR #164, opened 2023-01-09
cleanup: remove unnecessary flag initialization

boranx:
Type: Clean-up

This removes flag initialization that's performed during startup, as it's being checked by cobra already

Testing:
Check the output (against the previous version), you shouldn't be seeing any diff


mjlshen (approved):
/lgtm

PR #169, opened 2023-03-07
Move the tag creation to happen at instance creation time.

fahlmant:
This moves the step of creating tags into the same step that creates the EC2 instance and EBS volumes. This is required for the move to tag-based permission enforcement. Without this, instance will not be able to be created since the tags are required at creation time.


jharrington22 (approved):
/lgtm

PR #172, opened 2023-03-31
Allow consuming repos to obtain all URLs associated with egress failures

mjlshen:

This PR is a iteration in the vein of [OSD-15078](https://issues.redhat.com//browse/OSD-15078)

Currently, it is difficult for consuming repos to determine exactly
which egress URLs are blocked without manually doing regex parsing on an
array of failure strings that we return. This commit solves this by
adding an egressURL field with EgressURL() and GetEgressURLFailures()
methods so that consuming repos can more easily obtain this information.

## Checklist

- [ ] ~~I have made corresponding changes to the documentation~~

## Reviewer's Checklist

I updated the integration test so that it would print out the egress failures using the new functionality in this PR. Just running it should replicate my results.

## Logs 

The results of me adding this code block to the integration test (which only blocks quay.io:443)
```go
	for _, ef := range egressFailures {
		log.Printf("egress failure: %s", ef.EgressURL())
	}
```

```
❯ ./integration --region us-east-2
2023/03/31 15:16:31 using region us-east-2
2023/03/31 15:16:31 using AZ: us-east-2a
2023/03/31 15:16:31 created VPC: vpc-0cad5fd3990c440db
2023/03/31 15:16:31 enableDnsSupport is true for VPC: vpc-0cad5fd3990c440db
2023/03/31 15:16:31 enableDnsHostnames is true for VPC: vpc-0cad5fd3990c440db
2023/03/31 15:16:31 creating private subnet: 10.0.0.0/25
...a bunch of provisioning
2023/03/31 15:26:15 egress failure: quay.io:443
2023/03/31 15:26:15 ONV egress validation: Failure!
2023/03/31 15:26:15 deleting firewall arn:aws:network-firewall:us-east-2:429297027867:firewall/osd-network-verifier-firewall
...a bunch of deleting
```

### DIFF
--- a/pkg/output/output.go
+++ b/pkg/output/output.go
@@ -1,6 +1,7 @@
 package output
 
 import (
+	"errors"
 	"fmt"
 
 	handledErrors "github.com/openshift/osd-network-verifier/pkg/errors"
@@ -110,4 +111,21 @@ func (o *Output) Summary(debug bool) {
 // - errors as []error
 func (o *Output) Parse() ([]error, []error, []error) {
 	return o.failures, o.exceptions, o.errors
+}
+
+// GetEgressURLFailures returns only errors related to network egress failures.
+// Use the EgressURL() method to obtain the specific url for each error.
+func (o *Output) GetEgressURLFailures() []*handledErrors.GenericError {
+	egressErrs := []*handledErrors.GenericError{}
+
+	for _, err := range o.failures {
+		var nve *handledErrors.GenericError
+		if errors.As(err, &nve) {
+			if nve.EgressURL() != "" {
+				egressErrs = append(egressErrs, nve)
+			}
+		}
+	}
+
+	return egressErrs
 }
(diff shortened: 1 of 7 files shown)


pkg/output/output.go
```diff
@@ -111,3 +112,20 @@ func (o *Output) Summary(debug bool) {
 func (o *Output) Parse() ([]error, []error, []error) {
 	return o.failures, o.exceptions, o.errors
 }
+
+// GetEgressURLFailures returns only errors related to network egress failures.
+// Use the EgressURL() method to obtain the specific url for each error.
+func (o *Output) GetEgressURLFailures() []*handledErrors.GenericError {
```
mjlshen, line 118: Something to consider - does this provide value or should we just return a list of strings represnting the egressURLs

boranx: imo that makes sense. We can define better utils for output, to provide a smoother experience for callers
Also, it'd eliminate the need of parsing the output: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5256/diffs#5e0a821c90e90a1732a77f0021d9754deffdb39b_474_474 (at some point, we need to bump-up the version here too 😅 )

boranx (approved):
/lgtm
/approve

PR #174, opened 2023-04-03
Move Security Group tag creation to be part of the Security Group creation

fahlmant:
Similar to #169 

This moves the step of creating tags into the same step that creates the Security Group. This is required for the move to tag-based permission enforcement. Without this, the Security Group will not be able to be created since the tags are required at creation time.


jharrington22 (commented):
@fahlmant looks good, I assume that we can probably remove the `CreateTags` method from the client too?

fahlmant:
@jharrington22 yep, done

mjlshen (commented):
/lgtm

PR #179, opened 2023-04-20
Ensuring the SG cleanup in case of error when creating the instance

bdematte:
Bug fix for [OSD-16076](https://issues.redhat.com//browse/OSD-16076)
In case we got an error and we couldn't create the EC2 instance, we should still clean the SG up in case we created one so skipping the log analysis and the TerminateInstace but continuing the function to cleanup the SG. 
This is only a first part of the fix (even though it should be the most common case we are facing) - we still need to better handle the error case inside the 'CreateSecurityGroup' function itself. 

## Checklist

## Reviewer's Checklist


abyrne55 (approved):
/lgtm

Tested locally by creating an IAM user with a policy that allows everything the verifier needs, except for RunInstances. This triggered the newly-added error handler, which I've confirmed deletes the verifier's security group.

PR #181, opened 2023-04-25
Add of pubKey to debug instance to allow ssh

luis-falcon:
- Enhancement/Feature

This PR adds creates a parameter which allows passing of a public ssh key to the created debug instance in order to allow ssh access from our IP. In addition creates the extra security group rule to allow SSH into the instance and sets the timeout of the auto termination of the instance to 1 hour instead of 5 mins.

### DIFF
--- a/pkg/verifier/aws/entry_point.go
+++ b/pkg/verifier/aws/entry_point.go
@@ -3,6 +3,7 @@ package awsverifier
 import (
 	"encoding/base64"
 	"fmt"
+	"io/ioutil"
 	"strconv"
 
 	awsTools "github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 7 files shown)


luis-falcon:
> Hey Luis, i ran the PR locally and it looks like there's still a problem with the functionality:
> 
>     * the instance named 'osd-network-verifier' was terminated even though it was supposed to skip termination.

===Hey @lnguyen1401  this is for now (for testing done on purpose) on entry_point.go file on line 98 there is a TODO to put the time to 60mins however, i left it at 5 mins till all works. But if it is not enough time to test the connection then I can changed it. i will put 10 mins instead for the skip termination so the instance is up for a bit longer.

pkg/verifier/aws/entry_point.go
```diff
@@ -45,13 +49,44 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 	}
 
 	// Terminate a debug instance leftover from a previous run
+=======
+	//Terminate Debug Session
+>>>>>>> 82261bd (Add of pubKey to debug instance to allow ssh)
```
lnguyen1401, line 54: merge conflict

lnguyen1401 (changes requested):

luis-falcon:
Closing for a new one with less merge issues

PR #182, opened 2023-04-25
SDA-6670 | feat: add formating function to retrieve output logs

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6670

# What
Adds a formatted string output from logs instead of directly printing to stdout, clients can consume the raw string

# Why
Clients currently have no access to the debug logs

### DIFF
--- a/pkg/output/output.go
+++ b/pkg/output/output.go
@@ -7,6 +7,8 @@ import (
 	handledErrors "github.com/openshift/osd-network-verifier/pkg/errors"
 )
 
+const logFormat = " - %v\n"
+
 // Output can be used when showcasing validation results at the end of the execution.
 type Output struct {
 	// debugLogs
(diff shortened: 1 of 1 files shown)


pkg/output/output.go
```diff
@@ -53,64 +55,48 @@ func (o *Output) IsSuccessful() bool {
...
+	fmt.Print(o.Format(debug))
 }
 
 // Parse returns the data being stored on output
 // - failures as []error
 // - exceptions as []error
 // - errors as []error
-func (o *Output) Parse() ([]error, []error, []error) {
-	return o.failures, o.exceptions, o.errors
+func (o *Output) Parse() ([]error, []error, []error, []string) {
```
mjlshen, line 98: I would rather that we create a new function to retrieve debug logs from this output struct rather than continue to extend this `Parse()` function.

However, maybe I don't understand the problem - what is missing that this PR hopes to achieve? The description isn't super clear

mjlshen: Gotcha, yeah if `Format()` is all you need that would be great. If we find that we need `Parse()` I think we had better consider just making separate functions for each of:

```
	debugLogs []string
	// failures represents the failed validation tests
	failures []error
	// exceptions is to show edge cases where a verifier test couldn't be ran as expected
	exceptions []error
	// errors is collection of unhandled errors
	errors []error
```

or just making all of them public members in the struct

gdbranco: Cool, removed Parse changes for now. We can handle it later if necessary

PR #183, opened 2023-04-25
Add of pubKey to debug instance to allow ssh

luis-falcon:
- Enhancement/Feature

This PR adds creates a parameter which allows importing of a public ssh key to the created debug instance and cleans it up afterwards.

### DIFF
--- a/pkg/verifier/aws/entry_point.go
+++ b/pkg/verifier/aws/entry_point.go
@@ -3,6 +3,7 @@ package awsverifier
 import (
 	"encoding/base64"
 	"fmt"
+	"os"
 	"strconv"
 
 	awsTools "github.com/aws/aws-sdk-go-v2/aws"
@@ -14,8 +15,11 @@ import (
 	"github.com/openshift/osd-network-verifier/pkg/verifier"
 )
 
-// Base path of the config file
-const CONFIG_PATH_FSTRING string = "/app/build/config/%s.yaml"
+const (
+	// Base path of the config file
+	CONFIG_PATH_FSTRING = "/app/build/config/%s.yaml"
+	DEBUG_KEY_NAME      = "onv-debug-key"
+)
 
 // ValidateEgress performs validation process for egress
 // Basic workflow is:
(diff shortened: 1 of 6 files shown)


pkg/helpers/config/userdata.yaml
lnguyen1401, line 1: They Key passed in is still not assign to the instance. So when i ssh the connection doesn't establish

lnguyen1401 (commented):
The command is still not working properly:
- The public key passed in is not assigned to the instance.

fahlmant:
There should not be a public IP attached to this instance, or if there is, this should only ever be an internal, development option. When we run this in a customer account, we specifically do not run this in public subnets nor attach public IPs due to the risk. The instance should only have the ssh keypair attached. The networking part should be figured out seperately (a jumphost is probably the best way to do this)

pkg/helpers/config/userdata.yaml
luis-falcon: @lnguyen1401  i change code to attach a keypair now. Thanks

luis-falcon (commented):
@fahlmant  Ping for review. Thanks. :)

pkg/verifier/aws/entry_point.go
```diff
@@ -46,12 +50,49 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
...
 	// Terminate a debug instance leftover from a previous run
 	if vei.TerminateDebugInstance != "" {
+
+		//Terminate the debug instance
 		if err := a.AwsClient.TerminateEC2Instance(vei.Ctx, vei.TerminateDebugInstance); err != nil {
 			a.Output.AddError(err)
 		}
+
+		//Delete the debug keypair
+		_, err := a.AwsClient.DeleteKeyPair(vei.Ctx, &ec2.DeleteKeyPairInput{
```
fahlmant, line 60: This is called before ImportKeyPair is so this will likely error most of the time. Additionally, this should only be called if vei.ImportKeyPair is specified, otherwise this code will be ran regardless of if a key pair is specified or not

luis-falcon (commented):
@fahlmant  made some changes for the code. Thanks for the input.

pkg/verifier/aws/entry_point.go
luis-falcon: Made the changes to the code to check for the key pair existence before executing a deletion.

pkg/verifier/aws/entry_point.go
```diff
@@ -44,11 +48,59 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
...
+	if vei.ImportKeyPair != "" {
+		//Read the pubkey file content into a variable
+		PubKey, err := os.ReadFile(vei.ImportKeyPair)
+		debugPubKey = PubKey
+		if err != nil {
+			return a.Output.AddError(err)
+		}
+
+		//Import Keypair into aws keypairs to be attached later to the created debug instance
+		_, err = a.AwsClient.ImportKeyPair(vei.Ctx, &ec2.ImportKeyPairInput{
```
boranx, line 62: for consecutive runs against the same AWS account, the second and afterward executions fail due to the key is already imported 
We can check if the key is already exists, and clean-it-up before re-importing a new one with the same

```go
Using region: us-east-1
Summary:
printing out failures:
 - error performing ec2:ImportKeyPair: The keypair already exists
 ```

boranx (commented):
lgtm, tested locally against an AWS account, verified the key-pair is attached

I've left a minor comment regarding the UI/UX

pkg/verifier/aws/entry_point.go
luis-falcon: ah this is good catch Boran. I forgot to test this case. Will fix right now before merging

boranx: As discussed in https://redhat-internal.slack.com/archives/CMD1L32M9/p1683299251940989, we might prefer having the error message instead of cycling the old key

PR #184, opened 2023-04-27
Use available validator image instead of pulling for AWS

mjlshen:
Attempt to use an image that's already available in the VM rather than trying to pulling an image no matter what. This has the net result of always using the image that's already available for AWS.

This PR attempts to minimize the amount of network egress calls the AMI needs to make before running its checks. In OSD-16098, it was noticed that `docker pull` was timing out due to a missing proxy configuration. Instead of solving it directly (as importing a custom trust bundle for docker to trust is not trivial), this PR reorganizes the script such that it attempts to use the pre-pulled image first.

## Logs 

These three failures are "expected" for now and should be fixed by the new set of AMIs that come from [OSD-16171](https://issues.redhat.com//browse/OSD-16171)

System log:
```
USERDATA BEGIN
ff90d2cf6dd20607fefb21677668764a175de1103978b71b290f5e2a6c3450a9
Using IMAGE: ff90d2cf6dd2
Validating tagging.us-east-1.amazonaws.com:80
Validating registry.redhat.io:443
Validating quay.io:443
Validating sso.redhat.com:443
Validating sso.redhat.com:80
Validating pull.q1w2.quay.rhcloud.com:443
Validating openshift.org:443
Validating console.redhat.com:443
Validating console.redhat.com:80
Validating quay-registry.s3.amazonaws.com:443
Validating cert-api.access.redhat.com:443
Validating api.access.redhat.com:443
Validating infogw.api.openshift.com:443
Validating mirror.openshift.com:443
Validating api.openshift.com:443
Validating cart-rhcos-ci.s3.amazonaws.com:443
Validating registry.access.redhat.com:443
Validating ocm-quay-production-s3.s3.amazonaws.com:443
Validating quayio-production-s3.s3.amazonaws.com:443
Validating ec2.amazonaws.com:443
Validating iam.amazonaws.com:443
Validating route53.amazonaws.com:443
Validating sts.amazonaws.com:443
Validating events.amazonaws.com:443
Validating events.pagerduty.com:443
Validating api.deadmanssnitch.com:443
Validating nosnch.in:443
Validating inputs1.osdsecuritylogs.splunkcloud.com:9997
Validating http-inputs-osdsecuritylogs.splunkcloud.com:443
Validating observatorium.api.openshift.com:443
Validating ec2.us-east-1.amazonaws.com:443
Validating elasticloadbalancing.us-east-1.amazonaws.com:443
Validating servicequotas.us-east-1.amazonaws.com:443
Validating servicequotas.us-east-1.amazonaws.com:80
Validating tagging.us-east-1.amazonaws.com:443

Not all endpoints were reachable:
Unable to reach events.amazonaws.com:443 within specified timeout after 3 retries: Get "https://events.amazonaws.com": dial tcp: lookup events.amazonaws.com on 10.0.0.2:53: no such host
Unable to reach tagging.us-east-1.amazonaws.com:80 within specified timeout after 3 retries: Get "http://tagging.us-east-1.amazonaws.com": dial tcp 52.94.233.76:80: connect: connection refused
Unable to reach servicequotas.us-east-1.amazonaws.com:80 within specified timeout after 3 retries: Get "http://servicequotas.us-east-1.amazonaws.com": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
USERDATA END
```

```
❯ ~/git/openshift/osd-network-verifier/osd-network-verifier egress --subnet-id subnet-02537289237936c1c --security-group-id sg-094091624bada491b --region us-east-1
Using region: us-east-1
Created instance with ID: i-068fbf16f2e95c92b
Summary:
printing out failures:
 - egressURL error: events.amazonaws.com:443
 - egressURL error: tagging.us-east-1.amazonaws.com:80
 - egressURL error: servicequotas.us-east-1.amazonaws.com:80

Failure!
```


pkg/helpers/config/userdata.yaml
```diff
@@ -20,8 +20,8 @@ write_files:
       fi
       echo "${USERDATA_END}" >> /var/log/userdata-output
 runcmd:
-  - sudo service docker start 2>1 > /dev/null || echo "docker not started by systemctl"
-  - /run-container.sh
+  - sudo systemctl start docker 2>1 > /dev/null || echo "docker not started by systemctl"
+  - HTTP_PROXY=${HTTP_PROXY} HTTPS_PROXY=${HTTPS_PROXY} /run-container.sh
```
mjlshen, line 24: fyi this enables `sudo docker pull` in the body of the userdata to work correctly when `HTTP_PROXY` and `HTTPS_PROXY` are defined (was missing before), however it still will not work if the customer provides an additional trust bundle.

pkg/helpers/config/userdata.yaml
```diff
@@ -1,17 +1,17 @@
 #cloud-config
-repo_update: true
```
mjlshen, line 2: Updating at runtime just adds time which we are trying to minimize and is just "another thing" that could stop verifier from running and we already get the latest updates at build time. Let's focus on keeping a known good AMI and having it up-to-date?

pkg/verifier/aws/entry_point.go
```diff
@@ -61,6 +61,7 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		"VALIDATOR_START_VERIFIER": "VALIDATOR START",
 		"VALIDATOR_END_VERIFIER":   "VALIDATOR END",
 		"VALIDATOR_IMAGE":          networkValidatorImage,
+		"VALIDATOR_REPO":           networkValidatorRepo,
```
mjlshen, line 64: This wasn't being set in the past and must have slipped through :D

pkg/helpers/config/userdata.yaml
```diff
@@ -1,17 +1,17 @@
...
     permissions: 755
     content: |
       #!/bin/bash
-      # based on tls, set up docker run command.
       echo "${USERDATA_BEGIN}" >> /var/log/userdata-output
-      sudo docker pull ${VALIDATOR_IMAGE} || echo "Warning: could not pull the specified docker image, will try to use the prepulled one" >> /var/log/userdata-output
-      VALIDATOR_REFERENCE=`echo ${VALIDATOR_IMAGE} | cut -d : -f 1`
-      # Retrieving the latest image successfully pulled (either from the script, or prepulled in the AMI)
+      # Look for the image pre-pulled in the AMI, if not available, try to pull it
       IMAGE=`docker images ${VALIDATOR_REPO} -q  | head -n 2 | tail -n 1`
```
bdematte, line 9: should be VALIDATOR_REFERENCE (not VALIDATOR_REPO, this is a typo) - https://issues.redhat.com/browse/OSD-13343 
I started doing some testing on this, but as you are updating this part, would make sense to update it all together.

pkg/helpers/config/userdata.yaml
```diff
@@ -1,17 +1,17 @@
...
-repo_update: true
 write_files:
   - path: /run-container.sh
     permissions: 755
     content: |
       #!/bin/bash
-      # based on tls, set up docker run command.
       echo "${USERDATA_BEGIN}" >> /var/log/userdata-output
-      sudo docker pull ${VALIDATOR_IMAGE} || echo "Warning: could not pull the specified docker image, will try to use the prepulled one" >> /var/log/userdata-output
-      VALIDATOR_REFERENCE=`echo ${VALIDATOR_IMAGE} | cut -d : -f 1`
```
bdematte, line 11: as images are referenced by sha256 now, cut should be on @ instead of `:`

pkg/verifier/aws/entry_point.go
bdematte: This should be calculated through VALIDATOR_REFERENCE rather than duplicating entries.

pkg/verifier/gcp/gcp_verifier.go
```diff
@@ -16,10 +16,11 @@ import (
 	computev1 "google.golang.org/api/compute/v1"
 )
 
-var (
+const (
 	// TODO find a location for future docker images
-	networkValidatorImage string = "quay.io/app-sre/osd-network-verifier:v0.1.159-9a6e0eb"
```
bdematte, line 21: should be referenced by sha256

pkg/helpers/config/userdata.yaml
bdematte: should be done in the reverse order imho : be in a position to have regular deliveries of AMI (and pulling updating one easily) before removing the update part

pkg/verifier/gcp/gcp_verifier.go
mjlshen: I'm not going to update the GCP image here - it doesn't work for other reasons and don't want to muddy the waters

bdematte: then, just don't touch the GCP file :)

pkg/helpers/config/userdata.yaml
mjlshen: We have just been refreshing the package listing, there's no associated `repo_upgrade: true`

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html#user-data-cloud-init

pkg/verifier/gcp/gcp_verifier.go
bdematte: Adding a variable can be discussed - I am not a huge fan (mainly because the approach of constantly dupplicating element rather than limiting the amount of variable led to one of the current issues of having partial updates because of the duplication) but there are definitively pros and cons. 
However, the reasoning "it doesn't work for other reasons and don't want to muddy the waters" applies the same here :D . Anyway, this is detail, I will make a MR to aline GCP docker image.. 
For the additional variable topic, may be good to bring it to the team channel. I am not a huge fan of it, but best is to get opinion of others I would say

pkg/helpers/config/userdata.yaml
mjlshen: I'll raise this in a separate ticket - this change shouldn't be a part of this PR.

PR #188, opened 2023-05-11
Fix bash error causing osd-network-verifier timeouts pulling from quay.io

mjlshen:

Fixes a bug when osd-network-verifier times out due to being unable to `docker pull`, even when it is not necessary

```bash
      IMAGE=`docker images ${VALIDATOR_REPO} -q  | head -n 2 | tail -n 1`
      if [[ -n "${IMAGE}" ]]; then # should be -z for if $IMAGE is empty
        sudo docker pull ${VALIDATOR_IMAGE} >> /var/log/userdata-output
```

Also some improvements to the integration test, sometimes the test errors because the VPC can take a bit to initialize and cleaned up the help menu.

[OSD-16322](https://issues.redhat.com//browse/OSD-16322)

## Logs 

Original integration test help menu:

```
❯ ./integration -h                            
Usage of ./integration:
  -alsologtostderr
    	log to standard error as well as files
  -create-only
    	When specified, only create infrastructure and do not delete
  -delete-only
    	When specified, delete infrastructure in an idempotent fashion
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -log_dir string
    	If non-empty, write log files in this directory
  -log_link string
    	If non-empty, add symbolic links in this directory to the log files
  -logbuflevel int
    	Buffer log messages logged at this level or lower (-1 means don't buffer; 0 means buffer INFO only; ...). Has limited applicability on non-prod platforms.
  -logtostderr
    	log to standard error instead of files
  -profile string
    	AWS Profile
  -region string
    	AWS Region (default "us-east-1")
  -stderrthreshold value
    	logs at or above this threshold go to stderr (default 2)
  -v value
    	log level for V logs
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging

panic calling String method on zero glog.vModuleFlag for flag vmodule: runtime error: invalid memory address or nil pointer dereference
```

New:
```
❯ ./integration -h
Usage of ./integration:
  -create-only
    	When specified, only create infrastructure and do not delete
  -delete-only
    	When specified, delete infrastructure in an idempotent fashion
  -profile string
    	AWS Profile
  -region string
    	AWS Region (default "us-east-1")
```

After fixing this bug, the integration test passes as expected with:
```
❯ ./osd-network-verifier egress --subnet-id subnet-0b0d86c0835e9e905 --security-group-id sg-025aae27d87cbb34d                                                                                                
Using region: us-east-1
Created instance with ID: i-03f10708f4e1efef6
Summary:
printing out failures:
 - egressURL error: quay.io:443

Failure!
```

### DIFF
--- a/integration/main.go
+++ b/integration/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"os"
 	"time"
 
 	ocmlog "github.com/openshift-online/ocm-sdk-go/logging"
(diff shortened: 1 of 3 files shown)


lnguyen1401:
Tested against a cluster that block quay.io and works well !

PR #194, opened 2023-06-28
(OSD-16076) osd-network-verifier security group not being cleaned up

dakotalongRH:
Added a new function, CleanupSecurityGroup that is defered to always be called at the end.

[OSD-16076](https://issues.redhat.com//browse/OSD-16076): https://issues.redhat.com/browse/OSD-16076
There is a bug where SGs are sometimes not being cleaned up when network-verifier is ran.  This is likely because there are multiple places that network-verifier can error out where the SG cleanup would not have been triggered.

New function CleanupSecurityGroup was written to manually clean up the security groups.  It uses the 'defer' keyword in Go to always to execute at the _end_ of the function it's nested in.  In this case, CleanupSecurityGroup should always be called even if a previous line errors out.

- [X] I have commented my code, particularly in hard-to-understand areas

## Reviewer's Checklist

## Logs 

Logs from a local build that show when we pass an empty string as the SG and network-verifier creates one for us, it is still deleted.  Verified by seeing the SG created in AWS console and deleted after network-verifier runs.

`[dalong@dalong osd-network-verifier]$ ./osd-network-verifier egress --subnet-id subnet-0fc61bbebbe88c6fb --security-group-id "" --region us-east-1 --profile default
Using region: us-east-1
Created security group with ID: sg-0e8ba75fe5b8f73d9
Created instance with ID: i-09ab3d810b7721b3e
Summary:
All tests passed!
Success`

Logs from a local build that have an SG passed in (where we would not want to delete) run successfully.  These SG's will not be marked for deletion and the CleanupSecurityGroup function is never called.
`[dalong@dalong osd-network-verifier]$ ./osd-network-verifier egress --subnet-id subnet-0fc61bbebbe88c6fb --security-group-id sg-04d391ba59378ff3a --region us-east-1 --profile default
Using region: us-east-1
Created instance with ID: i-0e68f4bf6739e83a9
Summary:
All tests passed!
Success
`

### DIFF
--- a/pkg/verifier/aws/entry_point.go
+++ b/pkg/verifier/aws/entry_point.go
@@ -170,6 +170,11 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		keyPair:         vei.ImportKeyPair,
 	})
 
+	//If securitygroup was created by network-verifier, delete it as part of cleanup
+	if cleanupSecurityGroup {
+		defer CleanupSecurityGroup(vei, a)
+	}
+
 	if err != nil {
 		a.Output.AddError(err)
 	} else {
@@ -181,15 +186,6 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 			if err := a.AwsClient.TerminateEC2Instance(vei.Ctx, instanceID); err != nil {
 				a.Output.AddError(err)
 			}
-		}
-	}
-
-	if cleanupSecurityGroup {
-		_, err := a.AwsClient.DeleteSecurityGroup(vei.Ctx, &ec2.DeleteSecurityGroupInput{GroupId: awsTools.String(vei.AWS.SecurityGroupId)})
-		if err != nil {
-			a.Output.AddError(handledErrors.NewGenericError(err))
-			a.Output.AddException(handledErrors.NewGenericError(fmt.Errorf("unable to cleanup security group %s, please manually clean up", vei.AWS.SecurityGroupId)))
-
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/verifier/aws/entry_point.go
```diff
@@ -243,3 +239,14 @@ func (a *AwsVerifier) VerifyDns(vdi verifier.VerifyDnsInput) *output.Output {
 
 	return &a.Output
 }
+
+// Cleans up the security groups created by network-verifier
+func CleanupSecurityGroup(vei verifier.ValidateEgressInput, a *AwsVerifier) *output.Output {
+	_, err := a.AwsClient.DeleteSecurityGroup(vei.Ctx, &ec2.DeleteSecurityGroupInput{GroupId: awsTools.String(vei.AWS.SecurityGroupId)})
+	if err != nil {
```
mjlshen, line 246: Not exactly the same case, but it would be good to compare this with https://github.com/openshift/aws-vpce-operator/blob/main/controllers/vpcendpoint/cleanup.go#L133-L152 to not return an error if the error code from AWS is `InvalidGroup.NotFound`

dakotalongRH: I agree we should add checks for that but I think it's out of the scope of this card.  The specific issue here is there are multiple places we return with an error where the security groups would be left dangling.  The current implementation will guarantee that CleanupSecurityGroup will always be called regardless of erroring out earlier.

rafael-azevedo (approved):
/lgtm

pkg/verifier/aws/entry_point.go
```diff
@@ -170,6 +170,11 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		keyPair:         vei.ImportKeyPair,
 	})
 
+	//If securitygroup was created by network-verifier, delete it as part of cleanup
+	if cleanupSecurityGroup {
+		defer CleanupSecurityGroup(vei, a)
```
boranx, line 175: That makes a lot sense, thanks! 👍

PR #203, opened 2023-11-20
OSD-19187 allow multiple SGs to be passed to Network verifier

luis-falcon:
- Enhancement/Feature

This PR allows to pass multiple security groups to network verifier in order to attach multiple SGs to the network verifier instance

## Reviewer's Checklist

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -31,7 +31,7 @@ type egressConfig struct {
 	vpcSubnetID                string
 	cloudImageID               string
 	instanceType               string
-	securityGroupId            string
+	securityGroupIds           []string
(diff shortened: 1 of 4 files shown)


mjlshen:
OSD-19187

abyrne55 (changes requested):
Thanks for the PR! Unfortunately, per the card's done criteria, we can't delete `SecurityGroupId` right away. We will have to accept both `SecurityGroupId`  and `SecurityGroupIds` for at least a few versions so that we don't break backwards compatibility.

Of course, I'm open to arguments to the contrary :)

luis-falcon:
closing in favor of another PR that will better meet the done criteria.

PR #205, opened 2023-11-23
add security-groups-ids flag to the the command. allows multiple SGs …

luis-falcon:
Enhancement:

Add security-groups-ids flag to the the command. allows multiple SGs to be attached to network verifier instance

Still not tested. submit PR for initial Review.


cmd/egress/cmd.go
```diff
@@ -235,10 +238,11 @@ are set correctly before execution.
 	validateEgressCmd.Flags().BoolVar(&config.skipAWSInstanceTermination, "skip-termination", false, "(optional) Skip instance termination to allow further debugging")
 	validateEgressCmd.Flags().StringVar(&config.terminateDebugInstance, "terminate-debug", "", "(optional) Takes the debug instance ID and terminates it")
 	validateEgressCmd.Flags().StringVar(&config.importKeyPair, "import-keypair", "", "(optional) Takes the path to your public key used to connect to Debug Instance. Automatically skips Termination")
-
 	if err := validateEgressCmd.MarkFlagRequired("subnet-id"); err != nil {
 		validateEgressCmd.PrintErr(err)
 	}
 
+	//Mark securityGroupId and securityGroupsIDs flags as mutually exclusive (one or the other should be passed, not both).
+	validateEgressCmd.MarkFlagsMutuallyExclusive("security-group-id", "security-groups-ids")
```
boranx, line 246: should we have a check also in the library itself? E.g what happens if the caller passes both `securityGroupId` and `securityGroupsIDs ` to `createEC2InstanceInput{}`
One more q: what was the motivation behind introducing this to a new flag instead making `security-group-id` a list, Is it because of backward-compatibility purposes?

boranx (commented):
all good, I've left 2 small questions to clarify the context
Thanks for the contribution!

cmd/egress/cmd.go
luis-falcon: @boranx  first questions. this is handled by the "validateEgressCmd.MarkFlagsMutuallyExclusive("security-group-id", "security-groups-ids")" on the cmd.go file. this is a flag grouping method that makes sure when 1 of the flags is present the other one isnt. otherwise will output an error message asking you to use one of them not both.

second. I am of the same opinion. Originally, I had built the code completely replacing the single groupID by a list as you mentioned PR HERE( https://github.com/openshift/osd-network-verifier/pull/203.) but The done criteria of the ticket as stated by @abyrne55  is to deprecate the usage of the old flag (meaning have the new functionality in place but giving time for people to adapt I guess and keep back compatibility for a while). eventually I guess with the refactor of network verifier we will just have one flag as list.

boranx: Ack, then we don't consider library consumers (ppl who call pkg/ functions directly from their app, without using cmd.go) as they would need to refactor their app to use the second flag

PR #208, opened 2024-01-31
Added t3.micro as default fallback when non nitro instances are chose…

luis-falcon:
Added t3.micro as default fallback when non nitro instances are chose…

https://issues.redhat.com/browse/OSD-20512


abyrne55 (changes requested):
Made a few changes to make logging more consistent. Also reintroduced the  "set default if blank" bit of code, because external apps (not the CLI) will end up printing a nasty error otherwise. See [feedback.patch.txt](https://github.com/openshift/osd-network-verifier/files/14118268/feedback.patch.txt) (apply with `git am < feedback.patch.txt`)

abyrne55 (approved):
/lgtm

Tested this locally with `--instance-type p3.2xlarge` and confirmed the verifier quietly defaulted back to t3.micro. Thanks @luis-falcon!

mjlshen (approved):
/approve

PR #209, opened 2024-02-07
added a fix for the leaking security groups.

luis-falcon:
## Reviewer's Checklist


pkg/verifier/aws/entry_point.go
```diff
@@ -206,6 +207,43 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
...
+
+		//Getting a list of the SGs for the current VPC of our instance
+		var defaultSecurityGroupID = ""
+		describeSGOutput, err := a.AwsClient.DescribeSecurityGroups(vei.Ctx, &ec2.DescribeSecurityGroupsInput{
+			GroupNames: []string{},
+			Filters: []ec2Types.Filter{
+				{
+					Name:   awsTools.String("vpc-id"),
+					Values: []string{vei.AWS.VpcID},
+				},
```
mjlshen, line 222:
```suggestion
				{
					Name:   awsTools.String("vpc-id"),
					Values: []string{vei.AWS.VpcID},
				},
				{
					Name:   awsTools.String("group-name"),
					Values: []string{"default"},
				},
```
We should be able to combine the filters here so that we can delete the filtering loop underneath this

mjlshen (changes requested):
Looks good overall, noticed a small area of improvement :D

pkg/verifier/package_verifier.go
```diff
@@ -39,6 +39,7 @@ type AwsEgressConfig struct {
 	KmsKeyID         string
 	SecurityGroupId  string // Deprecated: prefer securityGroupIDs
 	SecurityGroupIDs []string
+	VpcID            string
```
abyrne55, line 42: Requiring this parameter is a non-trivial change to our API. Could we avoid having to ask the caller for this, e.g., by checking the VPC associated with the provided security group ID? Or we could note the VPC attached to the probe instance, perhaps?

abyrne55 (changes requested):

pkg/verifier/package_verifier.go
luis-falcon: This is what I am doing @abyrne55  this param I am not requesting from the user. I am populating it with a function GetVpcIdFromSubnetId() and filling it. however, we can also change it for a variable in the function scope.

pkg/clients/aws/aws.go
```diff
@@ -83,12 +83,17 @@ type EC2Client interface {
...
 	DeleteKeyPair(ctx context.Context, params *ec2.DeleteKeyPairInput, optFns ...func(*ec2.Options)) (*ec2.DeleteKeyPairOutput, error)
 	DescribeKeyPairs(ctx context.Context, params *ec2.DescribeKeyPairsInput, optFns ...func(*ec2.Options)) (*ec2.DescribeKeyPairsOutput, error)
+	ModifyInstanceAttribute(ctx context.Context, params *ec2.ModifyInstanceAttributeInput, optFns ...func(*ec2.Options)) (*ec2.ModifyInstanceAttributeOutput, error)
 }
 
 func (c *Client) DescribeKeyPairs(ctx context.Context, params *ec2.DescribeKeyPairsInput, optFns ...func(*ec2.Options)) (*ec2.DescribeKeyPairsOutput, error) {
 	return c.ec2Client.DescribeKeyPairs(ctx, params, optFns...)
 }
 
+func (c *Client) ModifyInstanceAttribute(ctx context.Context, params *ec2.ModifyInstanceAttributeInput, optFns ...func(*ec2.Options)) (*ec2.ModifyInstanceAttributeOutput, error) {
```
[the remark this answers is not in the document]
mjlshen: Wait...we don't have this in HCP is this a problem? https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.15/hypershift/sts_hcp_installer_permission_policy.json

luis-falcon (commented):
changes proposed by mshen implemented @abyrne55  let me know please read my answer to you comment and let me know if you prefer a variable or if we leave the vei.VpcId in the struct.

pkg/verifier/package_verifier.go
luis-falcon: Changed it to a regular variable after some thought.

mjlshen (approved):
/approve

PR #210, opened 2024-02-14
fix a bug introduced with the describe security Group  filters

luis-falcon:
fix a bug introduced where the describe security Group  filters were built with empty value for VpcID

make sure the VpcId always has a value. extracted from the if that executed only when we dont use a --security-group-id flag. Now we always have the VpcId which we need in other parts of the code.


fahlmant (approved):
/lgtm

PR #212, opened 2024-02-16
Add return whenever describeSG fails

luis-falcon:


pkg/verifier/aws/entry_point.go
```diff
@@ -229,7 +229,7 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 			},
 		})
 		if err != nil {
-			a.Output.AddError(err)
+			return a.Output.AddError(err)
```
AlexVulaj, line 232: This looks good to me, but I'll leave it to someone with more experience in the codebase to point out if returning earlier here would cause any issues.

PR #213, opened 2024-02-16
sanity check the describeSGOutput to avoid nil pointer erros during i…

luis-falcon:
sanity check the describeSGOutput to avoid nil pointer erros during integration tests. Also we get nice fail message. In addition I removed the line in filters for the SG "GroupNames: []string{}," causing a bug in the SDK to happen (Bug described here https://github.com/aws/aws-sdk/issues/402.

wraps the function to dettach SGs from instances into a sanity check for nil pointers to avoid errors when API does not respond or other issues.


rafael-azevedo (approved):
looks good

abyrne55:
/lgtm
/approve

Tested locally

PR #215, opened 2024-02-16
Cleanup security groups during integration test cleanup

AlexVulaj:
- Enhancement/Feature

During resource cleanup, integration tests fail to delete a VPC if there are dangling security groups attached. This change will make sure that those security groups are

## Logs 

These logs used to appear during cleanup:
```
panic: operation error EC2: DeleteVpc, https response error StatusCode: 400, RequestID: 3ee80c4a-badb-415a-9b95-f568cf795270, api error DependencyViolation: The vpc 'vpc-0fa551f9b466b36a6' has dependencies and cannot be deleted.
```
This change will fix that error.

### DIFF
--- a/integration/pkg/aws/cleanup.go
+++ b/integration/pkg/aws/cleanup.go
@@ -3,6 +3,8 @@ package aws
 import (
 	"context"
 	"errors"
+	"fmt"
+	ec2Types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/aws/smithy-go"
 	"log"
 	"time"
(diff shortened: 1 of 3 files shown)


integration/pkg/aws/cleanup.go
```diff
@@ -13,6 +15,46 @@ import (
...
+	if len(resp.SecurityGroups) == 0 {
+		log.Println("No SecurityGroups found - skipping cleanup")
+		return nil
+	}
+
+	for _, sg := range resp.SecurityGroups {
+		if _, err := id.ec2Api.DeleteSecurityGroup(ctx, &ec2.DeleteSecurityGroupInput{GroupId: sg.GroupId}); err != nil {
+			return err
+		}
+		log.Printf("deleted Security Group: %s", *sg.GroupId)
```
abyrne55, line 35: We should make this more noticable warning-level log message, e.g.,
```
WARN: Found a leaked osd-network-verifier security group. This is unexpected, as the verifier should fully clean up after itself. Cleaning up anyways...
```

integration/pkg/aws/cleanup.go
```diff
@@ -13,6 +15,47 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	if len(resp.SecurityGroups) == 0 {
+		log.Println("No SecurityGroups found - skipping cleanup")
+		return nil
+	}
+
+	log.Println("WARN: Found a leaked osd-network-verifier security group. This is unexpected, as the verifier should fully clean up after itself. Cleaning up anyways...")
```
AlexVulaj, line 31: @abyrne55 I added that message here, as this is the point at which we would have a non-zero number of security groups to clean up. This way, the message below still logs individual SGs that get deleted.

rafael-azevedo (approved):
/lgtm

PR #216, opened 2024-02-19
Allow http-proxy flag to allow domain names in the url not just IP address

luis-falcon:
Allow http-proxy flag to allow domain names in the url not just IP address

Closes https://issues.redhat.com/browse/OSD-20562


luis-falcon:
closing to fix module errors.

PR #218, opened 2024-02-20
Allow http-proxy flag to allow domain names in the url not just IP ad…

luis-falcon:
Allow http-proxy flag to allow domain names in the url not just IP address

Closes https://issues.redhat.com/browse/OSD-20562


pkg/verifier/aws/aws_verifier.go
```diff
@@ -521,16 +522,35 @@ func (a *AwsVerifier) CreateSecurityGroup(ctx context.Context, tags map[string]s
...
 	ipUrl, err := url.Parse(ipUrlStr)
 	if err != nil {
 		return nil, err
 	}
 
-	// Then attempt to extract IP address
-	ipAddr, err := netip.ParseAddr(ipUrl.Hostname())
+	//extract the hostName from the URL
+	hostValue := ipUrl.Hostname()
+	ipFromDomain := hostValue
```
AlexVulaj, line 534: nitpick - you could get away with only using one variable here if you wanted by doing `ipFromDomain := ipUrl.Hostname()` here, and then passing it into the `validate.Var` call instead of `hostValue`.

pkg/verifier/aws/aws_verifier.go
```diff
@@ -521,16 +522,35 @@ func (a *AwsVerifier) CreateSecurityGroup(ctx context.Context, tags map[string]s
...
+	err = validate.Var(hostValue, "fqdn")
+	if err != nil {
+		//do nothing in order to proceed to check if hostValue is an IP address
+	} else {
+		//if the hostValue is a fqdm set the ip to 0.0.0.0 in order to create an Outbound SG rule to all IPs
+		//Ref: OSD-20562
+		ipFromDomain = "0.0.0.0"
+	}
+
+	////Check if the hostname is an IP address
```
AlexVulaj, line 549: Extra comment lines can be removed here
```suggestion
	// Check if the hostname is an IP address
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -521,16 +522,35 @@ func (a *AwsVerifier) CreateSecurityGroup(ctx context.Context, tags map[string]s
 // ipPermissionFromURL generates an EC2 IpPermission (for use in sec. group rules) from a given IP address
 // -based http(s) URL (e.g. "https://10.0.8.1:8080") with the given human-readable description (ipPermDescription)
```
abyrne55, line 523:
```suggestion
// ipPermissionFromURL generates an EC2 IpPermission (for use in sec. group rules) from a given http(s) 
// URL (e.g., "http://10.0.8.1:8080" or "https://proxy.example.com:1234") with the given human-readable
// description (ipPermDescription)
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -521,16 +522,35 @@ func (a *AwsVerifier) CreateSecurityGroup(ctx context.Context, tags map[string]s
...
+
+	//check if the hostname is a fully qualified domain name
+	err = validate.Var(hostValue, "fqdn")
+	if err != nil {
+		//do nothing in order to proceed to check if hostValue is an IP address
+	} else {
+		//if the hostValue is a fqdm set the ip to 0.0.0.0 in order to create an Outbound SG rule to all IPs
+		//Ref: OSD-20562
+		ipFromDomain = "0.0.0.0"
+	}
```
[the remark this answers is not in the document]
luis-falcon: i like having the if != nil even if empty since It keeps a standard on the code. but I accept the  suggestion I changed.

pkg/verifier/aws/aws_verifier.go
```diff
@@ -521,16 +522,33 @@ func (a *AwsVerifier) CreateSecurityGroup(ctx context.Context, tags map[string]s
 // ipPermissionFromURL generates an EC2 IpPermission (for use in sec. group rules) from a given IP address
 // -based http(s) URL (e.g. "https://10.0.8.1:8080") with the given human-readable description (ipPermDescription)
 func ipPermissionFromURL(ipUrlStr string, ipPermDescription string) (*ec2Types.IpPermission, error) {
-	// First validate URL by parsing it
+
+	//validate URL by parsing it
 	ipUrl, err := url.Parse(ipUrlStr)
```
abyrne55, line 527: As this function is no longer dealing with just IP-address-based URLs, let's change the names of `ipUrlStr` and `ipUrl` to something like `allowedUrl[Str]` to avoid confusion

abyrne55 (changes requested):
Great work! Requested a few readability changes on top of those suggested by @AlexVulaj

pkg/verifier/aws/aws_verifier.go
luis-falcon: I will for the first couple. however since in the end what we want to be dealing with even when we get a fully qualified domain is to work with IPs in order to add them to SGs, i will leave the name of variables towards the end as they are.

pkg/verifier/aws/aws_verifier.go
luis-falcon: Great catch. Changed

pkg/verifier/aws/aws_verifier.go
luis-falcon: ohh you right. testing.

abyrne55 (changes requested):
Found a bug that's probably my fault, so I made the patch for you: [slashzero.patch.txt](https://github.com/openshift/osd-network-verifier/files/14365251/slashzero.patch.txt). Use `git am < slashzero.patch.txt` to apply. Basically if you create a rule with "0.0.0.0/32", it only allows access to the literal 0.0.0.0 address, not all IPv4 addresses as desired. So I added logic to switch the /32 to /0, and renamed some other variables to keep things consistent

PR #220, opened 2024-02-22
reverting the commits that broke tests for HCP clusters by modifying 

luis-falcon:
This PR reverts changes introduced to the network verifier. The original PRs that are being reverted allowed for the SGs to be deleted even if the instance termination took too long and timed out or failed (which before would leave SGs also on the CU's account). However those 2 PRs since they required the modifyInstanceAttribute permissions on the AWS account (which for Regular OSD was fine but not for Hypershift) it was breaking network-verifier tests for hypershift clusters. 

This PR Reverts those original changes until we have managed policies for modifyInstanceAttributes in Hypershift.


rafael-azevedo (approved):
/lgtm

abyrne55:
Mentioning for posterity: this PR rolled back #209

PR #221, opened 2024-02-26
enabled IMDSv2 service for the network verifier instance

luis-falcon:
https://issues.redhat.com/browse/OSD-20983

Enables instance Metadata service v2 on the ec2 instance created by the network verifier. The IMDSv2 introduces session-oriented requests, which adds defense in depth against unauthorized metadata access. This will not make big change on the network verifier instance at the moment since we do not request any metadata about the instance. But if at some point we do we can just call the same IMDSv1  IP url for it but now with an authentication token.


rafael-azevedo (approved):
/lgtm

PR #222, opened 2024-02-29
Updated egress lists documentationt o point to correct line for registering new platform types

AlexVulaj:
The target file in the diff has been changed such that [Line 46](https://github.com/openshift/osd-network-verifier/blob/main/pkg/helpers/helpers.go#L46) is no longer accurate for the location of platform types variables. They now live on [Line 94](https://github.com/openshift/osd-network-verifier/blob/main/pkg/helpers/helpers.go#L94).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Golden AMI provides the following:
 
 ### Egress Lists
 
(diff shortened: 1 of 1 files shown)


rafael-azevedo (approved):
/lgtm

PR #224, opened 2024-03-04
added --force-temp-security-group flag

luis-falcon:
#type of change
- Enhancement/Feature

https://issues.redhat.com/browse/OSD-20969

Added a --force-temp-security-group flag that will force the creation of a security group to attach to the instance whenever this flag is passed (whether we pass with flags other security groups or not).

## Checklist

Run network verifier with the following flags
1 ) "--force-temp-security-group " and "--security-group-id" (expected 2 SGs to be attached to instance, the temp and the passed. The created one to be cleaned at the end).).

2) "--force-temp-security-group " and "--security-group-ids" (expected all SGs passed and the newly created one to be attached to instance, and the created one to be cleaned at the end).

3) if all the 3 flags are passed an error since the security-group-id and security-group-ids flags are are mutually exclusive.

4) if no security group flags are passed, then the default is to create a new SG to attach to instance whether we pass the "--force-temp-security-group" flag or not.

5) Integration tests must pass.

## Reviewer's Checklist

Follow the 5 steps above to test.

Explain here how this PR will be tested by the reviewer: commands, dependencies, steps, etc.

Review the previous 5 steps

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -48,6 +48,7 @@ type egressConfig struct {
 	skipAWSInstanceTermination bool
 	terminateDebugInstance     string
 	importKeyPair              string
+	ForceTempSecurityGroup     bool
 }
 
 func getDefaultRegion(platformType string) string {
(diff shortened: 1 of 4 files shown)


abyrne55 (approved):
Tested locally and confirmed via AWS console that temporary security group is created, attached to probe instance, and deleted whenever `--force-temp-security-group` is passed, regardless of value of `--security-group-id[s]`. Also successfully ran integration test.

Nice work @luis-falcon. Please file a separate PR fulfilling AC "Docs and example code have been updated to reflect this new parameter."

PR #225, opened 2024-03-08
Update Docs and example of using new force-temp-security-group flag

luis-falcon:
## Type of Change
- Docs

Updates the Documentation and Gives an example of using the new --force-temp-security-group flag


rafael-azevedo (approved):
/lgtm

PR #226, opened 2024-03-17
fix leaking SG group issue

luis-falcon:
- Bug Fix

Closes https://issues.redhat.com/browse/OSD-20777

When the network verifier we do not pass a security group with flag --security-group-id we create a temporary one. Now before instance termination I detach this temp SG and replace it by the default SG for the instance which will allow the SG to be deleted whether the instance properly terminates or not (timeouts)

## Reviewer's Checklist
run the network verifier with no --security-group-id passed.
Go into aws console and watch the SG change during the final moments of the network verifier run.
NOTE: Might have to go into the instance page to view changes since the instance dashboard view sometimes do not refresh.


pkg/verifier/aws/entry_point.go
```diff
@@ -207,6 +208,45 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 
 	// Terminate the EC2 instance (unless user requests otherwise)
 	if !vei.SkipInstanceTermination {
+		//Replaced the SGs atached to the network-verifier-instance by the default SG in order to allow
```
AlexVulaj, line 211:
```suggestion
		//Replace the SGs attached to the network-verifier-instance by the default SG in order to allow
```

pkg/verifier/aws/entry_point.go
```diff
@@ -207,6 +208,45 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
...
+		if err != nil {
+			a.Output.AddError(err)
+		}
+
+		//Filtering SGs to get only the default SG ID finishing the iteration as soon as we find the default SG.
+		for _, SG := range describeSGOutput.SecurityGroups {
+			if *SG.GroupName == "default" {
+				defaultSecurityGroupID = *SG.GroupId
+			}
+		}
```
abyrne55, line 238: Did you mean to put a `break` here to finish the iteration early, as you state in the comment?

pkg/verifier/aws/entry_point.go
```diff
@@ -207,6 +208,45 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
...
+
+		//Replacing the SGs attach to instance by the default one. This is to clean the SGs created in case the instance
+		//termination times out
+		_, err = a.AwsClient.ModifyInstanceAttribute(vei.Ctx, &ec2.ModifyInstanceAttributeInput{
+			InstanceId: &instanceID,
+			Groups:     []string{defaultSecurityGroupID},
+		})
+		if err != nil {
+			a.Logger.Info(vei.Ctx, "Warning: Could not modify instance attribures. Proceeding with Instance Deletion without dettaching Security Groups")
```
abyrne55, line 247:
```suggestion
			a.Logger.Info(vei.Ctx, "Unable to detach instance from security group. Falling back to slower cloud resource cleanup method.")
			a.writeDebugLogs(vei.Ctx, fmt.Sprintf("Error encountered while trying to detach instance: %s.", err))
```

abyrne55 (changes requested):
Tested locally using AWS IAM profile that denies `ec2:ModifyInstanceAttributes`. Good work @luis-falcon. I've suggested a couple tweaks to enhance debug output and performance, but core biz logic is sound.

pkg/verifier/aws/entry_point.go
luis-falcon: I did but not needed anymore. it was a comment from my first try. The way is now the filter should return a unique SG anyways. Updating the comment

abyrne55 (approved):
Tested latest commit locally and observed updated error message and still-working functionality. Good stuff!

/lgtm
/approve

PR #235, opened 2024-05-15
Fix nil pointer error when DescribeSecurityGroups is not possible dur…

luis-falcon:
Fix nil pointer error when DescribeSecurityGroups is not possible during clean up of resources

-Bug fix

fixes issue with nil pointer when trying to iterate through a list of Security groups. We get this list with a DescribesecurityGroups call but if this fails the variable had a nil pointer.

To Fix I made a conditional to only execute the code detaching the SG from the instance only if we get a result from the DescribeSecurityGroup api Call.

## Reviewer's Checklist

- run the network verifier, run integration tests.

### DIFF
--- a/pkg/verifier/aws/entry_point.go
+++ b/pkg/verifier/aws/entry_point.go
@@ -217,7 +217,6 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 
 		//Getting a list of the SGs for the current VPC of our instance
 		var defaultSecurityGroupID = ""
-
 		describeSGOutput, err := a.AwsClient.DescribeSecurityGroups(vei.Ctx, &ec2.DescribeSecurityGroupsInput{
 			Filters: []ec2Types.Filter{
 				{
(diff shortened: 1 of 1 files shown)


abyrne55 (approved):
Tested and ensured reachability of 
1. "Unable to describe security groups" error when `ec2:DescribeSecurityGroups` is denied by IAM
2. "Unable to detach instance from security group" error when `ec2:ModifyInstanceAttribute` is denied by IAM
3. `ModifyInstanceAttribute` when all actions allowed by IAM

Note that I was only able to reach Case 1 when specifying an existing security group a (i.e., when the verifier is not asked to create a temporary security group). This is because the temporary security group verification process will fail if `ec2:DescribeSecurityGroups` is denied by IAM, preventing us from ever reaching any of the code touched by this PR. I believe OCM always uses `ForceTempSecurityGroup`, so it's strange that they're seeing this error. Perhaps customers have some very arcane IAM rules in place that are allowing the first `ec2:DescribeSecurityGroups` call to go through (during temp. SG creation) while denying later calls to the same API endpoint 🤷🏻

Regardless of the above caveat, approving because the bug is still reproducible in some cases, and this fix works in those cases. Nice work @luis-falcon 🙌🏻

PR #236, opened 2024-05-29
Bump golden ami ids for next release of the network verifier

luis-falcon:
-Release

What does this PR do?
Bump golden ami ids for next release of the network verifier

Bump the ami and ran integration tests according to SOP.


rafael-azevedo (approved):
/lgtm

PR #237, opened 2024-05-29
fixing double sha25 typo

luis-falcon:
fixing double sha25 typo


abyrne55:
Tested locally and verified new image is being used
/lgtm
/approve

PR #240, opened 2024-06-26
OSD-22892 (Part 1): Machine ID Lookup Table

abyrne55:

These commits cover the definition of the Probe interface's `GetMachineImageID()` method, as well as a reference implementation for the CurlJSONProbe. This method takes 3 string arguments (`platformType`, `cpuArch`, and `region`) and returns a "machine image ID," which is a cloud-platform-agnostic term for AMI ID. The `platformType` and `cpuArch` arguments should be one of the `helpers.Platform*` and `helpers.Arch*` enumerated string types, respectively, while `region` is more free-form. An example invocation might look like:
```go
imageID, err = myCurlJSONProbe.GetMachineImageID(helpers.PlatformAWS, helpers.ArchX86, "us-east-1")
```
On the backend, CurlJSONProbe's implementation of `GetMachineImageID()` is essentially a lookup table, i.e., a multi-level dictionary (in "machine-images.go") containing all the machine image IDs, plus a helper function with some lookup validation logic (e.g., to account for the fact that GCP machine image IDs are not region-specific like AMIs). The x86 AMI IDs included in this PR are the first "RHEL repackaged" images published to the verifier's golden AMI account: these are not expected to change often (maybe 1-3x per year). Ditto for the GCP machine image IDs, which just point to the publicly-available GCP RHEL images. ARM image IDs have placeholder values for now.

This PR intentionally does *not* cover integration of this new code into the verifier's critical path. That integration will be covered in a separate cherry-picked PR. As such, there isn't much to execute for this PR beyond the included unit tests. **When reviewing this PR, please focus on code style/structure.**

## Checklist


pkg/probes/curl_json/machine_images.go
```diff
@@ -0,0 +1,81 @@
...
+import "github.com/openshift/osd-network-verifier/pkg/helpers"
+
+// cloudMachineImageMap is a lookup table mapping VM image IDs to their
+// respective cloud platforms, CPU architectures, and cloud regions. To
+// access, reference cloudMachineImageMap[$CLOUD_PLATFORM][$CPU_ARCH][$REGION];
+// e.g., cloudMachineImageMap[helpers.PlatformAWS][helpers.ArchX86]["us-east-1"]
+// Note that GCP images are global/not region-scoped, so the region key will
+// always be "*"
+var cloudMachineImageMap = map[string]map[string]map[string]string{
+	helpers.PlatformAWS: {
```
AlexVulaj, line 12: I think this is fine for now, but I worry about how we'll handle Legacy vs. RHEL9 base images in the future. Would the idea be an entirely separate set of nested maps?

abyrne55: Yup! In a totally separate file/directory. You can see how I did it for the legacy probe [here](https://github.com/openshift/osd-network-verifier/pull/241/files#diff-1f92ce6064310e73a3611adb32e0f3779e3dbd7abe1c7750c05486910b795ae2).

rafael-azevedo (approved):
/lgtm

PR #241, opened 2024-06-26
OSD-22892 (Part 2): Legacy Probe

abyrne55:

These commits introduce `LegacyProbe`, which implements the `Probe` interface similarly to CurlJSONProbe. The goal of this Probe is to mimic the functionality of the verifier pre-experimental-probe (i.e., [v0.4.10](https://github.com/openshift/osd-network-verifier/releases/tag/v0.4.10)) as closely as possible ("bug-for-bug" 😉) while still adhering to the letter and spirit of the new Probe interface. The x86 AMI IDs included in this PR are the last "golden AMI" images published to the verifier's golden AMI account: hopefully, we don't need to update these much at all.

This PR intentionally does *not* cover integration of this new code into the verifier's critical path. That integration will be covered in a separate cherry-picked PR. As such, there isn't much to execute for this PR beyond the included unit tests, and even those will likely fail to run until #240 is merged. **When reviewing this PR, please focus on code style/structure.**

## Checklist


pkg/probes/legacy/legacy.go
```diff
@@ -0,0 +1,162 @@
...
+	"fmt"
+	"os"
+	"regexp"
+
+	handledErrors "github.com/openshift/osd-network-verifier/pkg/errors"
+	"github.com/openshift/osd-network-verifier/pkg/helpers"
+	"github.com/openshift/osd-network-verifier/pkg/output"
+)
+
+type LegacyProbe struct{}
```
AlexVulaj, line 15: nitpick - I personally prefer to not name types starting with the package name, since it's redundant to external callers. e.g. another package calling this would look like `legacy.LegacyProbe`.

abyrne55: I hear you; I did feel a twinge of ick while writing that exact reference in the (soon-to-be-PR'd) integration code. Do you have any suggestions of how to avoid?

AlexVulaj: My vote would be to just go with `type Probe struct{}`, and then external callers would do `legacy.Probe` which is still clear to me.

abyrne55: Might that conflict with [the Probe interface](https://github.com/openshift/osd-network-verifier/pull/240/files#diff-05c2aa716bb4ebe6725fdbf3d33668763dbec6abc29aad7881e053dc4c07cdf7) though? Probably not literally due to being in different packages (although `legacy` is a child of `probes`), but mentally? Not trying to be obstinate, just want to best understand the suggestion

pkg/verifier/gcp/gcp_verifier.go
```diff
@@ -259,7 +259,8 @@ func generateUserData(variables map[string]string) (string, error) {
 	variableMapper := func(varName string) string {
 		return variables[varName]
 	}
-	data := os.Expand(helpers.UserdataTemplate, variableMapper)
+	// TODO: REPLACE JUNK VALUE "helpers.UserdataTemplate" BELOW
+	data := os.Expand("helpers.UserdataTemplate", variableMapper)
```
abyrne55, line 263: Note: this change is make the compiler happy. GCP support is already broken, so this logical "non-fix" doesn't actually take anything away.

PR #245, opened 2024-06-27
OSD-22892 (Part 3): CurlJSONProbe Improvements

abyrne55:

These commits improve CurlJSONProbe's ability to validate input variables and "render" special userdata template variables. For example, the verifier's existing critical path sets duration-as-a-string values  (e.g., "1m3s" for 63 seconds) for the the `TIMEOUT` and `DELAY` userdata variables. Curl's `--max-time ${TIMEOUT}` flag doesn't accept such non-numeric values, so we've added a function called `normalizeSaneNonzeroDuration()` that converts durations-as-strings into floating-point quantities of seconds, does some sanity checking (i.e., 0 < `TIMEOUT` < 3hrs), and returns a curl-safe float-as-a-string (or, in the case of `DELAY`, a cloud-init-safe int-as-a-string). In a similar vein, we also now translate the `NOTLS` and `CACERT` variables into their curl-safe equivalents.

This PR also lays the groundwork for "tlsDisabled" URL support (i.e., HTTPS URLs that we'd like curl to not verify the TLS cert for), which was not included in the original experimental curl probe. We do this by using curl's "parser reset" flag ("--next" or "-:") to do this, but this has the unfortunate side effect of forcing us to re-pass most curl flags (except global flags and those irrelevant to HTTPS).

This PR intentionally does *not* cover integration of this new code into the verifier's critical path. That integration is covered in #246. As such, there isn't much to execute for this PR beyond the included unit tests. **When reviewing this PR, please focus on code style/structure.**

## Checklist

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


fahlmant (approved):
/lgtm

PR #247, opened 2024-07-08
Remove deprecated 'securityGroupId' property from createEC2InstanceInput

AlexVulaj:
Removes deprecated `securityGroupId` property from `createEc2InstanceInput` struct.

Depends on PRs to the three known downstream consumers to stop using the field:
osdctl: https://github.com/openshift/osdctl/pull/592 (merged)
CAD: https://github.com/openshift/configuration-anomaly-detection/pull/300 (merged)
CS: No uses detected

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -34,7 +34,6 @@ type egressConfig struct {
 	vpcSubnetID                string
 	cloudImageID               string
 	instanceType               string
-	securityGroupId            string // Deprecated: prefer securityGroupIDs
 	securityGroupIDs           []string
 	cloudTags                  map[string]string
 	debug                      bool
@@ -142,7 +141,6 @@ are set correctly before execution.
 				//Setup AWS Specific Configs
 				vei.AWS = verifier.AwsEgressConfig{
 					KmsKeyID:         config.kmsKeyID,
-					SecurityGroupId:  config.securityGroupId,
 					SecurityGroupIDs: config.securityGroupIDs,
 				}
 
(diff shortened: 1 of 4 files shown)


abyrne55 (changes requested):
Unfortunately, I think there's more that has to happen here. [Consumers of the verifier](https://github.com/search?q=org%3Aopenshift+AWS.SecurityGroupId+NOT+repo%3Aopenshift%2Fosd-network-verifier+&type=code) are still setting the deprecated `SecurityGroupId` field of the `AwsEgressConfig` type instead of the `SecurityGroupIDs` field, and currently this PR leaves that deprecated field silently unhandled AFAICT. We either have to:
* preserve `AwsEgressConfig.SecurityGroupId` and replace the compatibility shim removed by this PR with another shim somewhere else
* yank `AwsEgressConfig.SecurityGroupId`, which is a breaking API change 

I'm not terribly opposed to breaking API changes at the moment, considering the verifier's next release will likely bump the x- and/or y-stream version, but we might want to be careful about changing too much all at once.

AlexVulaj:
@abyrne55 good catch on these concerns. My take is that - considering this is still in "beta" technically and only has 3 known consumers, we should update those consumers to use the new field instead and then come back to this PR. Curious to hear your thoughts or any other maintainers'!

AlexVulaj:
@abyrne55 I've updated osdctl and CAD with PRs linked in the description here. I didn't find any uses in clusters service.

abyrne55 (changes requested):
Nice work with those consumer PRs. One more thing here: could you just remove the `securityGroupId` field of the `egressConfig` struct type on cmd/egress/cmd.go:37, as well as the line that sets up a CLI flag for it near the bottom of the same file?

PR #249, opened 2024-07-11
Update existing GCP network verifier framework

eth1030:
This PR is part of [OSD-24064](https://issues.redhat.com/browse/OSD-24064) to update the existing GCP network verifier framework.

Changes were made to the GCP network to:
- create a RHEL9 instance
- enable running a GCP startup-script
- integrate new probe interface
- add dummy probe
- make findUnreachableEndpoints similar to AWS version

This PR does not include testing for new additions; they will be added later. The original GCP network verifier code did not run on its own and does not run on any customer accounts currently. Changes to the probe or a new probe are needed for GCP to run the network verifier.

## Checklist

- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have made corresponding change to the default configuration files~~
- [ ] ~~I have added execution results to the PR's readme~~


pkg/probes/dummy/dummy.go
```diff
@@ -0,0 +1,38 @@
+package dummy
+
+import (
+	"github.com/openshift/osd-network-verifier/pkg/output"
+)
+
+type DummyProbe struct{}
```
AlexVulaj, line 7: Can we rename this to just `Probe`? The package name is `dummy` so external callers will already have to qualify it as `dummy.Probe`, whereas `dummy.DummyProbe` can be a bit redundant.

pkg/probes/dummy/dummy.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// GetStartingToken returns the string token used to signal the beginning of the probe's output
+func (prb DummyProbe) GetStartingToken() string { return startingToken }
+
+// GetEndingToken returns the string token used to signal the end of the probe's output
+func (prb DummyProbe) GetEndingToken() string { return endingToken }
+
+// GetMachineImageID returns the string ID of the VM image to be used for the probe instance
+func (prb DummyProbe) GetMachineImageID(platformType string, cpuArch string, region string) (string, error) {
```
AlexVulaj, line 21: My personal preference is to delete the names of unused function params (here, line 26, and line 37). That makes it extremely clear to anyone looking at the code that we are strictly providing the parameters to satisfy an interface, but we don't care about the values themselves..

pkg/verifier/gcp/entry_point.go
```diff
@@ -25,25 +26,32 @@ func (g *GcpVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		vei.InstanceType = "e2-standard-2"
 	}
 
+	vei.InstanceType = "e2-micro"
```
AlexVulaj, line 29: can we make `e2-micro` a constant here? similar to how we do for `aws/entry_point.go#L23`. That way we can reuse it everywhere (including tests) and only need to change it in one spot if we have to update this in the future.

pkg/verifier/gcp/gcp_verifier.go
```diff
@@ -145,57 +151,64 @@ func (g *GcpVerifier) createComputeServiceInstance(input createComputeServiceIns
 
 }
 
-func (g *GcpVerifier) findUnreachableEndpoints(projectID, zone, instanceName string) error {
+func (g *GcpVerifier) findUnreachableEndpoints(projectID, zone, instanceName string, probe probes.Probe) error {
+
 	// Compile the regular expressions once
-	reVerify := regexp.MustCompile(userdataEndVerifier)
-	reUnreachableErrors := regexp.MustCompile(`Unable to reach (\S+)`)
+	// reVerify := regexp.MustCompile(userdataEndVerifier)
```
AlexVulaj, line 157: Should we just blow this code away?

pkg/verifier/gcp/gcp_verifier.go
```diff
@@ -145,57 +151,64 @@ func (g *GcpVerifier) createComputeServiceInstance(input createComputeServiceIns
...
-			// Check for the specific string we output in the generated userdata file at the end to verify the userdata script has run
-			// It is possible we get EC2 console output, but the userdata script has not yet completed.
-			verifyMatch := reVerify.FindString(string(scriptOutput))
-			if len(verifyMatch) < 1 {
-				g.Logger.Debug(context.TODO(), "ComputeService console output contains data, but end of userdata script not seen, continuing to wait...")
+		if len(output.Contents) == 0 {
+			g.Logger.Debug(context.TODO(), "ComputeService console output not yet populated with data, continuing to wait...")
+			return false, nil
+		}
+		consoleOutput = string(output.Contents)
```
AlexVulaj, line 179: output.Contents is already a string, so we can remove the boxing here.
```suggestion
		consoleOutput = output.Contents
```

pkg/probes/dummy/dummy.go
AlexVulaj: My personal preference is to delete the names of unused function params (here, line 26, and line 37). That makes it extremely clear to anyone looking at the code that we are strictly providing the parameters to satisfy an interface, but we don't care about the values themselves.
```suggestion
func (prb DummyProbe) GetMachineImageID(string, string, string) (string, error) {
```

eth1030: Is there an alternative to this?
```
pkg/verifier/gcp/entry_point.go:58:14: cannot use dummy.Probe{} (value of type dummy.Probe) as probes.Probe value in assignment: dummy.Probe does not implement probes.Probe (wrong type for method GetMachineImageID)
		have GetMachineImageID(string, string, string) (string, error)
		want GetMachineImageID(string, "github.com/openshift/osd-network-verifier/pkg/data/cpu".Architecture, string) (string, error)

AlexVulaj: That error looks... odd. Almost like a bug in our linter. Let's dig into this a bit more.

PR #250, opened 2024-07-12
OSD-24208 (Part 1): Add `cpu.Architecture` type

abyrne55:

This PR replaces the `Arch*` string constants in the `helpers` package with a new package `cpu` and a new named struct type `Architecture`. This adds type safety to our (and our consumers') specification of CPU architectures, as most incorrect values for parameters of type  `Architecture` will trigger a compiler error. I'd like to do something very similar with our enumerated type for cloud platforms (`PlatformType`) very soon.

This PR only includes the bare minimum integration of this new code into the verifier's critical path. Deeper integration will be covered in a separate PR. **Please feel free to run a cursory manual test when reviewing this PR, but afterwards, please focus on code style/structure.**

## Checklist


pkg/data/cpu/cpu_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+func TestArchitecture_DefaultInstanceType(t *testing.T) {
+	tests := []struct {
+		name         string
+		arch         Architecture
+		platformType string
+		want         string
+		wantErr      bool
+	}{
+		{
```
AlexVulaj, line 26: Is there any way we could get a test case to make sure we throw an error on invalid arch?

abyrne55: See the test case on lines 123-131

AlexVulaj: I see those, but I still think we should have a test case to ensure that `DefaultInstanceType` catches and returns an error for that case. This would be getting into the block on lines 53-55 in `cpu.go`

abyrne55: Thanks for clarifying; added it

PR #255, opened 2024-07-18
Add a flag to provide a custom location for the egress URL list

AlexVulaj:
- Enhancement/Feature

Adds a new flag to allow a user to pass a specific egress URL list.

Resolves [OSD-22629](https://issues.redhat.com//browse/OSD-22629)

To test locally - simply run the verifier as usual but pass in `--egress-list-location` with either an external url or a local file path (absolute or relative to the working directory).

### DIFF
--- a/pkg/data/egress_lists/egress_lists.go
+++ b/pkg/data/egress_lists/egress_lists.go
@@ -105,7 +105,7 @@ type endpoint struct {
 	TLSDisabled bool   `yaml:"tlsDisabled"`
 }
 
-// endpoint list type (as it appears in the current YAML schema)
+// reachabilityConfig list type (as it appears in the current YAML schema)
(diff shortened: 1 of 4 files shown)


pkg/verifier/aws/entry_point.go
AlexVulaj, line 1: This diff looks like more than it is because the new `if` block is confusing github.

pkg/data/egress_lists/egress_lists.go
AlexVulaj, line 1: I think at some point we should rewrite all of these functions to return `reachabilityConfig` (exported of course) so we aren't passing around arbitrary yaml strings, especially since there's an expected structure we're looking for in data. If others want I'm happy to do it as part of this effort, but my vote would be to push this off for now.

rafael-azevedo (approved):
/lgtm

PR #257, opened 2024-07-19
Add fallback if we fail to convert github list to a curl string

AlexVulaj:
- Enhancement/Feature

## What does this PR do?
Adds a fallback to use the local list if we fail to parse the content in the github list after successfully fetching it.

Relates to [OSD-22628](https://issues.redhat.com//browse/OSD-22628)


rafael-azevedo (approved):
/lgtm

PR #258, opened 2024-07-23
allow custom dns query

datianshi:
## Checklist

## Reviewer's Checklist

### DIFF
--- a/pkg/verifier/aws/aws_verifier.go
+++ b/pkg/verifier/aws/aws_verifier.go
@@ -59,6 +59,16 @@ var defaultIpPermissions = []ec2Types.IpPermission{
 			},
 		},
 	},
+	{
+		FromPort:   awsTools.Int32(53),
+		ToPort:     awsTools.Int32(53),
+		IpProtocol: awsTools.String("udp"),
+		IpRanges: []ec2Types.IpRange{
+			{
+				CidrIp: awsTools.String("0.0.0.0/0"),
+			},
+		},
+	},
 }
 
 const (


abyrne55:
Hi @datianshi, thanks for the PR! My understanding of [these OSD/ROSA docs](https://docs.openshift.com/rosa/rosa_planning/rosa-cloud-expert-prereq-checklist.html#custom-dns) is that users must not use custom DNS resolvers during cluster installation (but they can set them up after cluster installation completes, if they wish). IOW, a cluster's bootstrap/control-plane nodes can only use "AmazonProvidedDNS" a.k.a. Route53 Resolver until cluster installation is complete.

osd-network-verifier's goals include simulating a "mid-cluster-install bootstrap/control-plane node" as closely as possible while still complying with OSD/ROSA docs. It would seem that this PR does not serve that goal, as it would give the verifier the power to do something (make queries to a custom DNS resolver) that the OSD/ROSA docs say a cluster's nodes cannot do — at least, not during cluster installation/in-flight checks, which is when the temporary security group that this PR modifies would be used.

I'll note that the network verifier can use existing security groups if specified using the `--security-group-ids` flag or `SecurityGroupIDs` field of the ValidateEgressInput struct. You could accomplish what appears to be the goal of this PR by creating a security group that allows egress to port 53/udp, and then run the verifier like the following:
```
./osd-network-verifier egress --subnet-id=[your-target-subnet-id] --security-group-ids=[the-port-53-security-group-id] --force-temp-security-group
```
(I added `--force-temp-security-group` because otherwise the verifier doesn't create any security groups when an existing one is specified)

Of course, I'm open to discussion 🙂. Please let me know if I'm misunderstanding the linked docs or the goals of this PR

datianshi:
@michaelryanmcneill

michaelryanmcneill:
@abyrne55 after today's discussion in the MCS call, our documentation is being updated to properly reflect that it _is_ supported to use custom DHCP options. As such, I think this fix is acceptable and would ensure we are catching any sort of DNS resolution related problems for customers operating with custom DHCP options.

abyrne55:
I'm running into strange behavior while testing this. The SG is created with the additional rule as expected, and curl runs and is able to connect to most (if not all) endpoints, but the endingToken is never printed to console. Will continue testing

joshbranham (approved):
/lgtm

abyrne55:
Thanks for the PR and for your patience @datianshi. This feature will be included in the next verifier release

PR #259, opened 2024-07-24
Integrate new probe interface into GCP framework

eth1030:
Part of [OSD-24065](https://issues.redhat.com/browse/OSD-24065), this PR integrates new probe interface into GCP framework.

- added a startup-script analogous to the AWS userdata-script
- copy the curl json probe as gcp curl json probe, to make changes but not effect probe used by AWS
- go embed the startup-script.sh instead of userdata-template.yaml
- escape os.expand() by adding extra userDataVariables in GetExpandedUserData function because os.expand deletes unknown variables
- run new gcp_curl_json_probe instead of dummy probe
- added functionality to fetch url list from github

## Checklist

## Reviewer's Checklist

set environment variable GCP_PROJECT_ID to the project ID of the VPC
```
export GCP_PROJECT_ID=<your-VPC-project-ID>
```
run verifier  optionally with `--debug` flag
```
./osd-network-verifier egress --platform=gcp-classic --subnet-id <your-subnet-id> --vpc-name <your-vpc-name>
```

## Logs 
Expected output: verifier is able to log any blocked egresses
```
Using Project ID <project-ID>
Created instance with ID: verifier-<rand-num>
Applying labels
Successfully applied labels 
ComputeService Instance: verifier-<rand-num> RUNNING
Summary:
printing out failures:
 - egressURL error: https://events.pagerduty.com:443 (Empty reply from server)
 - egressURL error: https://events.amazonaws.com:443 (Could not resolve host: events.amazonaws.com)

Failure!
```
If client breaks or loses connection with instance, instance should self-delete within default or specified time


pkg/probes/gcp_curl/machine_images.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/openshift/osd-network-verifier/pkg/helpers"
+)
+
+// cloudMachineImageMap is a lookup table mapping VM image IDs to their
+// respective cloud platforms, CPU architectures, and cloud regions. To
+// access, reference cloudMachineImageMap[$CLOUD_PLATFORM][$CPU_ARCH][$REGION];
+// e.g., cloudMachineImageMap[helpers.PlatformAWS][cpu.ArchX86]["us-east-1"]
+// Note that GCP images are global/not region-scoped, so the region key will
+// always be "*"
+var cloudMachineImageMap = map[string]map[cpu.Architecture]map[string]string{
```
AlexVulaj, line 14: Is this just a place holder, or can GCP actually use AMIs?

pkg/probes/gcp_curl/machine_images.go
AlexVulaj, line 1: I think this entire file is unnecessary, as it's the same as the other `machine_images.go`.

AlexVulaj:
I'd really like to find a way to combine more of the common functionality here. For example, `pkg/probes/gcp_curl/curl_json_result.go`, `pkg/probes/gcp_curl/gcp_curl_json.go`, and `pkg/probes/gcp_curl/machine_images.go` are identical to their AWS counterparts.

Ideally the same probe should be used independently of AWS or GCP, perhaps with some configuration that can be passed in via parameters from the `entrypoint` files. (i.e. `userDataTemplate` and `userDataVariables`).

pkg/probes/legacy/legacy.go
```diff
@@ -79,7 +79,7 @@ func (lgp Probe) GetMachineImageID(platformType string, cpuArch cpu.Architecture
 // variables listed in the template's "network-verifier-required-variables" directive, or if
 // values *are* provided for variables that must be set to a certain value for the probe to
 // function correctly (presetUserDataVariables) -- this function will fill-in those values for you.
-func (lgp Probe) GetExpandedUserData(userDataVariables map[string]string) (string, error) {
+func (lgp Probe) GetExpandedUserData(userDataVariables map[string]string, template string) (string, error) {
```
AlexVulaj, line 82: I know it looks weird, but this is an established way in Go to signify that "Hey, I accept a parameter here to satisfy an interface, but I don't actually use it at all".

```suggestion
func (lgp Probe) GetExpandedUserData(userDataVariables map[string]string, _ string) (string, error) {
```

AlexVulaj:
/approve

Holding off on an lgtm as I don't have time to test this at the moment, but all looks good.

pkg/verifier/aws/entry_point.go
```diff
@@ -18,6 +19,9 @@ import (
 	"github.com/openshift/osd-network-verifier/pkg/verifier"
 )
 
+//go:embedd userdata-template.yaml
```
AlexVulaj, line 22: Typo here in `embed`
```suggestion
//go:embed userdata-template.yaml
```

eth1030:
After meeting with @AlexVulaj I propose a new PR where we groom the GCP egress URL list because `events.amazonaws.com` does not appear to be an actual endpoint. When running curl on my local device I get:
```
curl events.amazonaws.com
curl: (6) Could not resolve host: events.amazonaws.com
```
running the network verifier results in:
```
./osd-network-verifier egress --platform gcp-classic --subnet-id default --vpc-name default
Using Project ID emhammon
Created instance with ID: verifier-134
Applying labels
Successfully applied labels 
ComputeService Instance: verifier-134 RUNNING
Summary:
printing out failures:
 - egressURL error: https://events.amazonaws.com:443 (Could not resolve host: events.amazonaws.com)

Failure!
```
I think the network verifier is working correctly in identifying blocked egress, and the egress URL list for GCP should be groomed.

pkg/probes/curl/curl_json.go
```diff
@@ -81,8 +77,8 @@ func (clp Probe) GetMachineImageID(platformType string, cpuArch cpu.Architecture
 // variables listed in the template's "network-verifier-required-variables" directive, or if
 // values *are* provided for variables that must be set to a certain value for the probe to
 // function correctly (presetUserDataVariables) -- this function will fill-in those values for you.
-func (clp Probe) GetExpandedUserData(userDataVariables map[string]string) (string, error) {
-	// Extract required variables specified in template (if any)
+func (clp Probe) GetExpandedUserData(userDataVariables map[string]string, userDataTemplate string) (string, error) {
+	// Extract required variables specified in template (if any) based off platformType
```
AlexVulaj, line 81: Does the behavior here actually depend on `platformType`?

pkg/probes/package_probes.go
```diff
@@ -9,6 +9,6 @@ type Probe interface {
 	GetMachineImageID(platformType string, cpuArch cpu.Architecture, region string) (string, error)
 	GetStartingToken() string
 	GetEndingToken() string
-	GetExpandedUserData(map[string]string) (string, error)
+	GetExpandedUserData(map[string]string, string) (string, error)
```
AlexVulaj, line 12: Minor - this will help us remember what these values are supposed to be in the future if we ever implement a new probe.
```suggestion
	GetExpandedUserData(userDataVariables map[string]string, userDataTemplate string) (string, error)
```

pkg/verifier/gcp/entry_point.go
```diff
@@ -23,40 +29,77 @@ const (
...
 	g.Logger.Debug(vei.Ctx, "Using configured timeout of %s for each egress request", vei.Timeout.String())
 	//default gcp machine e2
 	if vei.InstanceType == "" {
 		vei.InstanceType = DEFAULT_INSTANCE_TYPE
 	}
+	if vei.Timeout == 0 {
+		vei.Timeout = DEFAULT_TIMEOUT
+	}
 
+	vei.Timeout = DEFAULT_TIMEOUT
```
AlexVulaj, line 46: This section looks a little confusing - it looks like `vei.Timeout` will always get set to `DEFAULT_TIMEOUT` no matter what. What's the intention here for setting that timeout?

pkg/verifier/gcp/entry_point.go
```diff
@@ -23,40 +29,77 @@ const (
...
+			if err != nil {
+				return g.Output.AddError(err)
+			}
+		}
+	}
+
+	// Generate the userData file
+	// As expand replaces all ${var} (using empty string for unknown ones), adding the env variables used in userdata.yaml
+	// os.Expand() deletes unknown variables, so we need to add them back in
 	userDataVariables := map[string]string{
```
AlexVulaj, line 86: My understanding is that these variables are eventually used to fill in `startup-script.sh`. If my understanding is correct, I think we can clean up a bunch of these variables that aren't used in there.

pkg/probes/curl/curl_json.go
eth1030: It does not! I have removed part of that comment.

pkg/verifier/gcp/entry_point.go
eth1030: Changed to `if vei.Timeout <= 0` and only setting timeout there based on those conditions.

pkg/verifier/gcp/entry_point.go
eth1030: Removed `AWS_REGION` because it was unused. Also added to the GCP startup-script to set proxy environment variables specified by `HTTP_PROXY` `HTTPS_PROXY` and encode specified `CACERT` in /proxy.pem and make executable. Although file is created every time a GCP instance is started, it is only used if `CACERT` is not empty (the curl probe adds it to `CURLOPTS` to be run )

pkg/probes/legacy/legacy.go
```diff
@@ -79,7 +79,7 @@ func (lgp Probe) GetMachineImageID(platformType string, cpuArch cpu.Architecture
 // variables listed in the template's "network-verifier-required-variables" directive, or if
 // values *are* provided for variables that must be set to a certain value for the probe to
 // function correctly (presetUserDataVariables) -- this function will fill-in those values for you.
-func (lgp Probe) GetExpandedUserData(userDataVariables map[string]string) (string, error) {
+func (lgp Probe) GetExpandedUserData(userDataVariables map[string]string, _ string) (string, error) {
```
abyrne55, line 82: Please add a comment to the doc string above noting why this probe doesn't need that last parameter

pkg/probes/curl/curl_json.go
```diff
@@ -81,7 +77,7 @@ func (clp Probe) GetMachineImageID(platformType string, cpuArch cpu.Architecture
 // variables listed in the template's "network-verifier-required-variables" directive, or if
 // values *are* provided for variables that must be set to a certain value for the probe to
 // function correctly (presetUserDataVariables) -- this function will fill-in those values for you.
-func (clp Probe) GetExpandedUserData(userDataVariables map[string]string) (string, error) {
+func (clp Probe) GetExpandedUserData(userDataVariables map[string]string, userDataTemplate string) (string, error) {
```
abyrne55, line 80: Please update docstring with new param

pkg/probes/legacy/legacy_test.go
```diff
@@ -213,7 +213,7 @@ func TestLegacyProbe_GetExpandedUserData(t *testing.T) {
 
 			prb := Probe{}
 			// First check if function is returning an error
-			got, err := prb.GetExpandedUserData(tt.userDataVariables)
+			got, err := prb.GetExpandedUserData(tt.userDataVariables, userDataTemplate)
```
abyrne55, line 216: Consider passing an empty string here instead of userdataTemplate to reduce potential confusion

pkg/verifier/gcp/entry_point.go
```diff
@@ -1,19 +1,26 @@
...
+	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
 	"github.com/openshift/osd-network-verifier/pkg/verifier"
 )
 
+//go:embed startup-script.sh
+var startupScript string
+
 const (
-	cloudImageIDDefault   = "rhel-9-v20240703"
+	DEFAULT_CLOUDIMAGEID  = "rhel-9-v20240703"
```
abyrne55, line 21: This is already defined inside the probe (machine_images.go)

pkg/verifier/gcp/entry_point.go
```diff
@@ -23,52 +30,86 @@ const (
...
+		"HTTPS_PROXY":      vei.Proxy.HttpsProxy,
+		"CACERT":           base64.StdEncoding.EncodeToString([]byte(vei.Proxy.Cacert)),
+		"NOTLS":            strconv.FormatBool(vei.Proxy.NoTls),
+		"DELAY":            "5",
+		"URLS":             egressListStr,
+		"TLSDISABLED_URLS": tlsDisabledEgressListStr,
+		"ret":              "${ret}",
+		"?":                "$?",
+		"array[@]":         "${array[@]}",
+		"value":            "$value",
```
abyrne55, line 99: Please add a comment explaining why this is necessary

pkg/verifier/gcp/entry_point.go
```diff
@@ -23,52 +30,86 @@ const (
...
+	userData, err := vei.Probe.GetExpandedUserData(userDataVariables, startupScript)
 	if err != nil {
 		return g.Output.AddError(err)
 	}
-
 	g.Logger.Debug(vei.Ctx, "Generated userdata script:\n---\n%s\n---", userData)
 
 	if vei.CloudImageID == "" {
-		vei.CloudImageID = cloudImageIDDefault
+		vei.CloudImageID = DEFAULT_CLOUDIMAGEID
```
abyrne55, line 109: Use `probe.GetMachineImageID()` here instead of a global constant. See also my above comment on that constant

pkg/verifier/gcp/gcp_verifier_functions_test.go
```diff
@@ -0,0 +1,81 @@
...
+		// TODO: Add test cases.
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := get_unreachable_endpoints(tt.args.consoleOutput, tt.args.probe); (err != nil) != tt.wantErr {
+				t.Errorf("get_unreachable_endpoints() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
```
abyrne55, line 81:
```suggestion
```
It's understandable to skip tests here for now, but no need to leave the boilerplate code in the meantime

pkg/verifier/gcp/gcp_verifier_functions_test.go
```diff
@@ -0,0 +1,81 @@
+package gcpverifier
+
+import (
+	"testing"
+
+	"github.com/openshift/osd-network-verifier/pkg/probes"
+	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
+)
+
+func Test_get_tokens(t *testing.T) {
```
abyrne55, line 10: Nice!

abyrne55 (changes requested):
This is awesome work @eth1030! I suggested some changes inline. In general, you'll want to explain some of the less-obvious workarounds you (very adeptly) implemented to get this working

pkg/verifier/aws/userdata-template.yaml
abyrne55, line 1: It would appear that two separate copies of this file now exist, which poses a maintainability challenge

pkg/verifier/aws/entry_point.go
```diff
@@ -177,7 +181,7 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		userDataVariables["DELAY"] = "60"
 	}
 
-	unencodedUserData, err := vei.Probe.GetExpandedUserData(userDataVariables)
+	unencodedUserData, err := vei.Probe.GetExpandedUserData(userDataVariables, userDataTemplate)
```
abyrne55, line 184: So this has us providing the same value of userDataTemplate to every probe, yeah? My first thought is that different probes need to use different user-data, but then I saw that your LegacyProbe just ignores the userDataTemplate param. So it works, but it's not immediately clear to somebody scanning through the code why this would work. IOW, this needs comments at the very least, but let's have a larger discussion about this first

pkg/verifier/aws/userdata-template.yaml
eth1030: Unfortunately go embed can only handle embedding files from the same directory, a solution is to take in a `platformType` parameter in `getExpandedUserData` preemptively embed both `userdata-template` and `startup-script` in the curl probe and then use one of them depending on the `platformType`. Any thoughts about this?

eth1030: Currently the `userdata-template` in the curl directory is only used for testing.

pkg/probes/curl/curl_json.go
```diff
@@ -26,6 +26,9 @@ type Probe struct{}
 //go:embed userdata-template.yaml
 var userDataTemplate string
 
+//go:embed startup-script.sh
+var startupScript string
```
abyrne55, line 30:
```suggestion
//go:embed systemd-template.sh
var systemdTemplate string
```
Renaming to make more consistent with userdata-template.yaml. More broadly, I'd like to rebrand your approach as the "systemd curl probe" vs. the "GCP curl probe" so that things don't get confusing if/when we try to use systemd on AWS in the future

pkg/probes/curl/curl_json.go
```diff
@@ -82,6 +85,13 @@ func (clp Probe) GetMachineImageID(platformType string, cpuArch cpu.Architecture
 // values *are* provided for variables that must be set to a certain value for the probe to
 // function correctly (presetUserDataVariables) -- this function will fill-in those values for you.
 func (clp Probe) GetExpandedUserData(userDataVariables map[string]string) (string, error) {
+	// Check if GCP instance is being used, if so use startup-script instead of userdata-template
+	// Only GCP should have USE_GCP_STARTUPSCRIPT userDataVariable
+	// Serves as a workaround for identifying platform type without a new probe interface
+	if userDataVariables["USE_GCP_STARTUPSCRIPT"] == "true" {
+		userDataTemplate = startupScript
+	}
```
abyrne55, line 93:
```suggestion
	// Use systemd to run curl (instead of cloud-init) if requested. Useful for
	// platforms that don't include cloud-init in their OS images (e.g., GCP)
	if userDataVariables["USE_SYSTEMD"] == "true" {
		userDataTemplate = systemdTemplate
	}
```
See above "rebrand" comment

pkg/probes/curl/machine_images.go
abyrne55, line 1: Just curious: is there a benefit to pinning to a specific image version? Does it still work if we just pass "rhel-9" instead of "rhel-9-v20240709"?

pkg/verifier/gcp/entry_point.go
```diff
@@ -23,52 +26,93 @@ const (
...
+	// Set timeout to default if not specified
+	if vei.Timeout <= 0 {
+		vei.Timeout = DEFAULT_TIMEOUT
+	}
+	g.Logger.Debug(vei.Ctx, "configured a %s timeout for each egress request", vei.Timeout)
+
+	// Set instance type to default if not specified and validate it
 	if vei.InstanceType == "" {
 		vei.InstanceType = DEFAULT_INSTANCE_TYPE
 	}
```
abyrne55, line 51:
```suggestion
	// Set instance type to default if not specified and validate it
	if vei.InstanceType == "" {
		vei.InstanceType, err = vei.CPUArchitecture.DefaultInstanceType(helpers.PlatformGCP)
		if err != nil {
			return g.Output.AddError(err)
		}
		g.Logger.Debug(vei.Ctx, fmt.Sprintf("defaulted to instance type %s", vei.InstanceType))
	}
```
Use `vei.CPUArchitecture.DefaultInstanceType()` instead of a `DEFAULT_INSTANCE_TYPE` constant in order to enable ARM-friendliness. We also probably need to do a little work on lining up the "user specified a bad instance type; what do?" logic with how it's done on the AWS side, but this works okay for now.

pkg/verifier/gcp/gcp_verifier_functions.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/openshift/osd-network-verifier/pkg/helpers"
+	"github.com/openshift/osd-network-verifier/pkg/probes"
+)
+
+func get_tokens(consoleOutput string, probe probes.Probe) bool {
```
abyrne55, line 11: Please add a docstring for this function

pkg/verifier/gcp/gcp_verifier.go
```diff
@@ -92,6 +92,11 @@ func (g *GcpVerifier) createComputeServiceInstance(input createComputeServiceIns
 			{
 				Name:       input.networkName,
 				Subnetwork: input.vpcSubnetID,
+				AccessConfigs: []*computev1.AccessConfig{
+					{
+						Name: "External NAT",
+					},
+				},
```
abyrne55, line 99: Not a blocker, but we should discuss this further (or at least put a TODO comment). It looks like this assigns a public IP to the instance, which was something we were asked to stop doing on the AWS side of things long ago. Please correct me if I'm misunderstanding GCP access configs though

abyrne55 (changes requested):
This is really, really good work and very close to merging.

pkg/probes/curl/machine_images.go
eth1030: It does not work if we pass in "rhel-9"

pkg/verifier/gcp/gcp_verifier.go
eth1030: Not 100% on this but https://pkg.go.dev/google.golang.org/api@v0.114.0/compute/v1#AccessConfig says 
```
	// AccessConfigs: An array of configurations for this interface.
	// Currently, only one access config, ONE_TO_ONE_NAT, is supported. If
	// there are no accessConfigs specified, then this instance will have no
	// external internet access.
```
which from what I have experienced would block egress as well. However, that is not to say there are no workarounds.

abyrne55: Ah thanks for clarifying! Sounds like this is spot-on then. Could you add a comment along the lines of "// Must specify an AccessConfig in order to enable VM internet access"?

pkg/verifier/gcp/entry_point.go
```diff
@@ -6,14 +6,16 @@ import (
...
 	"github.com/openshift/osd-network-verifier/pkg/output"
-	"github.com/openshift/osd-network-verifier/pkg/probes/dummy"
+	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
 	"github.com/openshift/osd-network-verifier/pkg/verifier"
 )
 
 const (
-	cloudImageIDDefault   = "rhel-9-v20240703"
-	DEFAULT_INSTANCE_TYPE = "e2-micro"
+	DEFAULT_TIMEOUT = 5
```
abyrne55, line 18:
```suggestion
	DEFAULT_TIMEOUT = 5 * time.Second
```
You'll also have to import `"time"` (if your IDE doesn't add the import for you automatically)

pkg/probes/curl/curl_json.go
```diff
@@ -82,6 +85,13 @@ func (clp Probe) GetMachineImageID(platformType string, cpuArch cpu.Architecture
 // values *are* provided for variables that must be set to a certain value for the probe to
 // function correctly (presetUserDataVariables) -- this function will fill-in those values for you.
 func (clp Probe) GetExpandedUserData(userDataVariables map[string]string) (string, error) {
+	// Check if GCP instance is being used, if so use startup-script instead of userdata-template
+	// Only GCP should have USE_GCP_STARTUPSCRIPT userDataVariable
+	// Serves as a workaround for identifying platform type without a new probe interface
```
abyrne55, line 90:
```suggestion
	// Use systemd to run curl (instead of cloud-init) if requested. Useful for
	// platforms that don't include cloud-init in their OS images (e.g., GCP)
```

abyrne55 (changes requested):
Only nitpicks in this one. After you fix those, feel free to post in the Aurora channel asking the team to test the code in their GCP accounts

pkg/verifier/gcp/gcp_verifier_functions.go
```diff
@@ -0,0 +1,44 @@
...
+	"strings"
+
+	"github.com/openshift/osd-network-verifier/pkg/helpers"
+	"github.com/openshift/osd-network-verifier/pkg/probes"
+)
+
+// function that tests probe order logic that is part of findUnreachableEndpoints in gcp_verifier.go
+// get_tokens checks for the presence of startingToken and endingToken in the consoleOutput
+// probe outsput should be between startingToken and endingToken
+func get_tokens(consoleOutput string, probe probes.Probe) bool {
```
AlexVulaj, line 14: nitpick here - use camelcase for function names in Go, i.e. `getTokens`

PR #260, opened 2024-08-01
Tune polling intervals for an overall faster `ValidateEgress()`

abyrne55:
This PR speeds up `ValidateEgress()` but adjusting how often we're checking for serial console output (i.e., "the `PollImmediate()` loop") and if the EC2 instance has finished terminating. This means that we'll spend less time unnecessarily waiting if, e.g., an instance finishes terminating after 17 seconds, as the previous config would only have checked once at 15s and again at 30s, whereas this new config will check at 3s, 6s, 9s... 15s, and 18s. 

## Checklist

- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- [x] I have tested the functionality against aws, it doesn't cause any regression

Keep an eye out for previously unforeseen race conditions or API ratelimiting (unlikely given the still-long intervals). Otherwise, please test this PR as you would any other verifier PR.

### DIFF
--- a/pkg/clients/aws/aws.go
+++ b/pkg/clients/aws/aws.go
@@ -155,9 +155,13 @@ func (c *Client) TerminateEC2Instance(ctx context.Context, instanceID string) er
 		return handledErrors.NewGenericError(err)
 	}
 
-	// Wait up to 5 minutes for the instance to be terminated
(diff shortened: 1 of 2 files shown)


rafael-azevedo (approved):
/lgtm

PR #262, opened 2024-08-05
Provide CA Certificates via cloud-init's ca_certs module

AlexVulaj:
- Bug

Currently, the curl flag we're using to provide a ca cert will _only_ use the provided cert - we want to use both the supplied SSL cert and RHEL's OEM trust bundle. This change moves away from the curl flag and instead uses the `ca_certs` module of cloud-init.

Resolves [OSD-24353](https://issues.redhat.com//browse/OSD-24353)

## Checklist

See [OSD-24353](https://issues.redhat.com//browse/OSD-24353) for setting up a transparent proxy. I used https://gitlab.cee.redhat.com/travi/osd-proxy-infra for setting up a non-transparent proxy.

I ran these new changes against
- a transparent proxy
- a non-transparent proxy
- a split-cert transparent proxy

In all situations, I saw the expected results. Elaborating more on the split-cert situation, I set up a transparent proxy using mitmproxy with `--ignore-hosts registry.redhat.io:443` in the service file. When I ran network verifier without passing a capath, _all_ URLs failed *_EXCEPT_* registry.redhat.io

## Logs 

With this change, the userdata template expands to something like the following:
```
#cloud-config

ca_certs:
  trusted:
    - |-
      -----BEGIN CERTIFICATE-----
      THE CERTIFICATE
      -----END CERTIFICATE-----
runcmd:
  - systemctl mask --now serial-getty@ttyS0.service
  - dmesg -D
  - echo "NV_CURLJSON_BEGIN" >/dev/ttyS0
  - export http_proxy= https_proxy=
  - curl --capath /etc/pki/tls/certs/ --retry 3 --retry-connrefused -t B -Z -s -I -m ${TIMEOUT} -w "%{stderr}${LINE_PREFIX}%{json}\n" ${CURLOPT} ${URLS} --proto =http,https,telnet ${TLSDISABLED_URLS_RENDERED} 2>/dev/ttyS0
  - echo "NV_CURLJSON_END" >/dev/ttyS0
power_state:
  delay: 2
  mode: poweroff
  message: Auto-terminating instance due to timeout
  timeout: 300
```


pkg/verifier/aws/entry_point.go
```diff
@@ -165,7 +165,7 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 		"TIMEOUT":          vei.Timeout.String(),
 		"HTTP_PROXY":       vei.Proxy.HttpProxy,
 		"HTTPS_PROXY":      vei.Proxy.HttpsProxy,
-		"CACERT":           base64.StdEncoding.EncodeToString([]byte(vei.Proxy.Cacert)),
```
abyrne55, line 168: We'll need to add a change to `legacy.Probe()` as it's still expecting a b64-encoded string

pkg/probes/curl/curl_json.go
```diff
@@ -104,17 +105,31 @@ func (clp Probe) GetExpandedUserData(userDataVariables map[string]string) (strin
...
+				},
+			},
+		}
+
+		cloudInitYamlBytes, cloudInitMarshalErr := yaml.Marshal(&cloudInit)
+		if cloudInitMarshalErr != nil {
+			return "", fmt.Errorf("unable to create cloud init config: %w", cloudInitMarshalErr)
+		}
+
+		userDataVariables["CACERT_RENDERED"] = strings.TrimSpace(string(cloudInitYamlBytes))
```
abyrne55, line 132: Neat approach, and smart way of avoiding complicated YAML escaping

abyrne55 (changes requested):

pkg/verifier/aws/entry_point.go
AlexVulaj: Good catch here. Looking at the `userdata-template.yaml` for the legacy probe, I see this line:
```
echo "${CACERT}" | base64 --decode > /proxy.pem
```
What do you think about just changing that to
```
echo "${CACERT}" > /proxy.pem
```

pkg/probes/legacy/userdata-template.yaml
```diff
@@ -15,7 +15,7 @@ write_files:
       fi
       echo "Using IMAGE: $IMAGE" >> /var/log/userdata-output
       if [[ "${CACERT}" != "" ]]; then
-        echo "${CACERT}" | base64 --decode > /proxy.pem
+        echo "${CACERT}" > /proxy.pem
```
AlexVulaj, line 18: @abyrne55 from your other comment - this should take care of things.

abyrne55: This works as long as there's no double-quotes (or other surprise characters) in $CACERT... which is a bit big of an assumption for a user-provided input. I'd suggest reversing this change and instead encoding the $CACERT input into base64 within this probe's `GetExpandedUserData()`, or adding some input sanitization logic. Clunky, I know, but the goal of the legacy probe is to emulate the verifier's old behavior as closely as possible — which sometimes means holding our noses.

pkg/verifier/aws/entry_point.go
abyrne55: See [my above reply](https://github.com/openshift/osd-network-verifier/pull/262/files#r1704582179)

pkg/probes/legacy/userdata-template.yaml
AlexVulaj: So to be clear - we'll keep the encoding in `entry_point` but just have the new curl probe decode it back? I'm cool with that!

pkg/probes/curl/userdata-template.yaml
```diff
@@ -6,7 +6,7 @@ runcmd:
   - dmesg -D
   - echo "${USERDATA_BEGIN}" >/dev/ttyS0
   - export http_proxy=${HTTP_PROXY} https_proxy=${HTTPS_PROXY}
-  - curl --retry 3 --retry-connrefused -t B -Z -s -I -m ${TIMEOUT} -w "%{stderr}${LINE_PREFIX}%{json}\n" ${CURLOPT} ${URLS} --proto =http,https,telnet ${TLSDISABLED_URLS_RENDERED} 2>/dev/ttyS0
+  - curl --capath /etc/pki/tls/certs/ --retry 3 --retry-connrefused -t B -Z -s -I -m ${TIMEOUT} -w "%{stderr}${LINE_PREFIX}%{json}\n" ${CURLOPT} ${URLS} --proto =http,https,telnet ${TLSDISABLED_URLS_RENDERED} 2>/dev/ttyS0
```
abyrne55, line 9:
```suggestion
  - curl --capath /etc/pki/tls/certs/ --proxy-capath /etc/pki/tls/certs/ --retry 3 --retry-connrefused -t B -Z -s -I -m ${TIMEOUT} -w "%{stderr}${LINE_PREFIX}%{json}\n" ${CURLOPT} ${URLS} --proto =http,https,telnet ${TLSDISABLED_URLS_RENDERED} 2>/dev/ttyS0
```
Necessary for proper operation behind non-transparent forward proxies

abyrne55 (changes requested):
Very good stuff. I re-ran tests behind a transparent proxy to great success. I was then able to tweak my terraform testing scripts enough to get a non-transparent proxy working, and that's when I realized that curl does some [extra SSL certificate checks](https://curl.se/docs/manpage.html#--proxy-capath) when non-transparent proxies are specified (e.g., via `osd-network-verifier egress --https-proxy="https://...`). So the suggestion below just tells curl to check the same trust store dir for proxy SSL certs as it does for "unproxied" (or transparently proxied) SSL certs.

abyrne55:
Tested locally. Nice work!

/lgtm
/approve

PR #275, opened 2024-10-03
OSD-25868: Adds `aws-hosted-cp` as a valid Platform name

joshbranham:

This extends the Platform structs to accept `aws-hosted-cp` as a valid value for looking up AWSHCP `ByName()`. This value is used internally still and blocks integrating the updated verifier.

## Checklist

## Reviewer's Checklist

### DIFF
--- a/pkg/data/cloud/platform_test.go
+++ b/pkg/data/cloud/platform_test.go
@@ -143,6 +143,10 @@ func TestByName(t *testing.T) {
 			name: "invalid name",
 			want: Platform{},
 		},
+		{
+			name: "",
+			want: Platform{},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/data/cloud/platform_test.go
```diff
@@ -123,6 +123,10 @@ func TestByName(t *testing.T) {
 			name: "aws-hcp",
 			want: AWSHCP,
 		},
```
abyrne55, line 125: Can we also add a test case for an empty string for `name`? Should return an empty `Platform{}`

abyrne55 (changes requested):
Good stuff, just requested an additional test case

pkg/data/cloud/platform_test.go
joshbranham: Good call, that failed so added a fix for that scenario as well.

abyrne55 (approved):
Tested locally with `./osd-network-verifier egress --platform=aws-hosted-cp` and confirmed new alias selects the aws-hcp egress list as expected. Also tested `--platform=""` and confirmed error.

Nice work, @joshbranham!

/lgtm

PR #279, opened 2024-10-09
(OSD-25668) Update documentation to include new Platform interface

dakotalongRH:
Update documentation to include the new platform type interface.

ype of change
- Docs

https://issues.redhat.com/browse/OSD-25668

## Checklist

- [X] I have made corresponding changes to the documentation
- [X] I have tested the functionality against gcp / aws, it doesn't cause any regression

## Reviewer's Checklist

### DIFF
--- a/docs/aws/aws.md
+++ b/docs/aws/aws.md
@@ -196,3 +196,24 @@ See the egress golang examples above, and replace the line starting with `out :=
 ```go
 out := cli.VerifyDns(context.TODO(), "vpcID")
 ```
+
+## Interface ##
+
+### Platform ###
+The platform interface is used to inform network-verifier of the platform type it is running on. It is used to represent specific Platform types and how they map to their respective platform. It contains an array of length containing valid names for each platform type.
+```
+type Platform struct {
+	// names holds 3 unique lowercase names of the Platform (e.g., "aws"). We use a fixed-
+	// size array so that this struct remains comparable. Any of the 3 values can be used to refer
+	// to this specific Platform via Platform.ByName(), but only the first (element
+	// 0) element will be the "preferred name" returned by Platform.String()
+	names [3]string
+}
+```
+
+Currently network-verifier supports three implementations for AWS.
+- AWSClassic
+- AWSHCP
+- AWSHCPZeroEgress
+
+Network-verifier uses these supported platform types to determine information such as which egress verification list, machine type, and cpu type to use.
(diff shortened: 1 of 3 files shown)


docs/aws/aws.md
```diff
@@ -196,3 +196,24 @@ See the egress golang examples above, and replace the line starting with `out :=
 ```go
 out := cli.VerifyDns(context.TODO(), "vpcID")
 ```
+
+## Interface ##
```
joshbranham, line 200: I think defining this in the main `README` or in something like `docs/platform.md` once would be easier and less duplication.

docs/aws/aws.md
```diff
@@ -196,3 +196,24 @@ See the egress golang examples above, and replace the line starting with `out :=
...
+	// to this specific Platform via Platform.ByName(), but only the first (element
+	// 0) element will be the "preferred name" returned by Platform.String()
+	names [3]string
+}
+```
+
+Currently network-verifier supports three implementations for AWS.
+- AWSClassic
+- AWSHCP
+- AWSHCPZeroEgress
```
joshbranham, line 217: Along with the above advice of consolidating, maybe a table for each `AWS` and `GCP` with the implementation names listed?

docs/aws/aws.md
```diff
@@ -196,3 +196,24 @@ See the egress golang examples above, and replace the line starting with `out :=
 ```go
 out := cli.VerifyDns(context.TODO(), "vpcID")
 ```
+
+## Interface ##
+
+### Platform ###
+The platform interface is used to inform network-verifier of the platform type it is running on. It is used to represent specific Platform types and how they map to their respective platform. It contains an array of length containing valid names for each platform type.
+```
```
joshbranham, line 204: One suggestion on wording here

> "The `Platform` struct type is used to inform network-verifier of the type of OSD platform it is running on (AWSClassic, GCPClassic, etc). This helps map the various aliases we support to identify a singular platform, for example AWSClassic can be referenced by the aliases `aws-classic` and `aws`.

I am still learning all these components but this is how I am led to understand it, so if I mispoke feel free to let me know and adjust 👍

docs/aws/aws.md
dakotalongRH: Agreed, moving to main README

docs/aws/aws.md
dakotalongRH: I think this would make sense if we had more implementations for each, but since GCP only has the one and AWS three I think the bulleted list is okay.  I'm still open to changing it

joshbranham: Yeah no strong opinion here, list seems good

joshbranham (commented):
/lgtm

PR #282, opened 2024-10-28
OSD-25966: Fix issue with `--egress-list-location` CLI flag

joshbranham:
## What does this PR do? / Related Issues / Jira

The `--egress-list-location` flag provided by the CLI was being ignored in the parsing logic that constructs URLS from given input YAML. This primarily ensures that if the `ValidateEgressInput` has a populated `EgressListYaml` value, we parse it and use it. I also moved the flow control around a bit for the decision making about where to source the data from to make it a bit easier to follow.

I wanted to refactor this further and consolidate the behavior between both AWS and GCP, but opted to tackle that in a follow on so it was easier to review the bug in isolation.

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have tested the functionality against gcp / aws, it doesn't cause any regression
- [x] I have added execution results to the PR's readme

## Reviewer's Checklist

- [ ] (This needs to be done after technical review) I've run the branch on my local, verified that the functionality is ok

## How to test this PR locally / Special Instructions

I ran this locally with the CLI to confirm it works with and without an `--egress-list-location` flag:

**With**
```
> ./osd-network-verifier egress --profile devaccount --subnet-id subnet-0bab965a3722b750d --no-proxy google.com --region us-west-2 --debug --egress-list-location=https://raw.githubusercontent.com/openshift/osd-network-verifier/refs/heads/main/pkg/data/egress_lists/aws-classic.yaml
Using region: us-west-2
Using external egress list from https://raw.githubusercontent.com/openshift/osd-network-verifier/refs/heads/main/pkg/data/egress_lists/aws-classic.yaml
```

**Without**
```
> ./osd-network-verifier egress --profile devaccount --subnet-id subnet-0bab965a3722b750d --no-proxy google.com --region us-west-2 --debug
Using region: us-west-2
configured a 5s timeout for each egress request
defaulted to x86 CPU architecture
defaulted to instance type t3.micro
defaulted to machine image ami-042d4c3472784287c
Using egress URL list from https://api.github.com/repos/openshift/osd-network-verifier/contents/pkg/data/egress_lists/aws-classic.yaml?ref=main at SHA c31b52add1381fa61e5c0446370c608e68feeb58
```

PR #284, opened 2024-11-01
OSD-26330: Refactor and add tests for `egress_lists` package

joshbranham:
When working on a bugfix in https://github.com/openshift/osd-network-verifier/pull/282, I wanted to consolidate duplicate code and clean up the control flow further for generating egress lists. This PR does just that, by first extracting the existing behavior into a common type, and adding tests for the various scenarios. A couple of notes:

* I did not add validation for invalid `PlatformType` as was setup before, since the caller should be using the various exported `vars` for `PlatformType` and not manually constructing there own alias type with custom `name` values.
* Opted for a small interface to abstract the GitHub API call to fetch contents
* Moved the various Getters in `egress_lists` to function receivers on the `Generator` struct. These functions are not used anywhere in the `openshift` organization outside of this project.
* Cleaned up some variables and comments

See below for output of running for both AWS and GCP:

### AWS

```
> ./osd-network-verifier egress --profile devaccount --subnet-id subnet-0bab965a3722b750d --region us-west-2        
Using region: us-west-2
Using egress URL list from https://api.github.com/repos/openshift/osd-network-verifier/contents/pkg/data/egress_lists/aws-classic.yaml?ref=main at SHA c31b52add1381fa61e5c0446370c608e68feeb58
Created security group with ID: sg-03f26bb00f4a69df7
Created instance with ID: i-0e1d53172c1849c21
Deleting instance with ID: i-0e1d53172c1849c21
Deleting security group with ID: sg-03f26bb00f4a69df7
Summary:
All tests passed!
Success
```
### GCP

**Note: The GCP failure is expected as the egress list is still being validated**

```
> GCP_PROJECT_ID=jbranham ./osd-network-verifier egress --platform gcp --vpc-name verifier-test-sg5b9-network --subnet-id verifier-test-sg5b9-worker-subnet        
Using Project ID jbranham
Using egress URL list from https://api.github.com/repos/openshift/osd-network-verifier/contents/pkg/data/egress_lists/gcp-classic.yaml?ref=main at SHA 92b7c45af595e5a23006d560222a57a83a0e61e3
Created instance with ID: verifier-1263
Applying labels
Successfully applied labels 
ComputeService Instance: verifier-1263 RUNNING
Summary:
printing out failures:
 - egressURL error: https://events.amazonaws.com:443 (Could not resolve host: events.amazonaws.com)

Failure!
```

## Checklist

## Reviewer's Checklist

You can perform a basic test locally by running the verifier and passing a custom list via the CLI

### DIFF
--- a/pkg/data/egress_lists/egress_lists.go
+++ b/pkg/data/egress_lists/egress_lists.go
@@ -1,21 +1,14 @@
 package egress_lists
 
-// TRANSITIONAL IMPLEMENTATION (UNSTABLE API)
-// This module currently provides very basic fetching of egress lists stored within the binary
-// in legacy probe/golden-AMI format. Most of its current logic & structs were borrowed from
-// osd-network-verifier-golden-ami/build/bin/network-validator.go with very little validation.
-// OSD-22628 will significantly change this module. Consider this internal API unstable until
-// that card is complete.
-
 import (
 	"context"
 	_ "embed"
 	"fmt"
+	"gopkg.in/yaml.v3"
 	"os"
 
 	"github.com/google/go-github/v63/github"
-	"gopkg.in/yaml.v3"
-
+	"github.com/openshift-online/ocm-sdk-go/logging"
 	"github.com/openshift/osd-network-verifier/pkg/data/cloud"
 )
 
@@ -31,12 +24,65 @@ var templateGCPClassic string
 //go:embed aws-hcp-zeroegress.yaml
 var templateAWSHCPZeroEgress string
 
-func GetLocalEgressList(platformType cloud.Platform) (string, error) {
-	if !platformType.IsValid() {
-		fmt.Printf("platform type %s is invalid", platformType)
+type githubReposClient interface {
+	GetContents(ctx context.Context, owner, repo, path string, opts *github.RepositoryContentGetOptions) (fileContent *github.RepositoryContent, directoryContent []*github.RepositoryContent, resp *github.Response, err error)
+}
+
+// Generator provides a mechanism to generate egress lists for a given platform and set of variables
+type Generator struct {
+	// PlatformType represents the cloud and type of platform we are generating egress lists for
+	PlatformType cloud.Platform
+
+	// Variables is a map of string:string used to replace templated values in canned egress lists
+	Variables map[string]string
+
+	logger            logging.Logger
+	githubReposClient githubReposClient
+}
+
+func NewGenerator(platformType cloud.Platform, variables map[string]string, logger logging.Logger) *Generator {
+	return &Generator{
+		PlatformType:      platformType,
+		Variables:         variables,
+		logger:            logger,
+		githubReposClient: github.NewClient(nil).Repositories,
+	}
+}
+
+// GenerateEgressLists takes an optional egressListYaml as input, and then attempts to return generated EgressLists
+// in the following order:
+// - If a populated egressListYaml is passed, use that
+// - Otherwise, try to get the values from GitHub, and if that fails
+// - Fallback to the local yaml embedded in this package
+func (g *Generator) GenerateEgressLists(ctx context.Context, egressListYaml string) (string, string, error) {
+	if egressListYaml != "" {
+		return g.EgressListToString(egressListYaml, g.Variables)
 	}
 
-	switch platformType {
+	egressResponse, err := g.GetGithubEgressList(ctx)
+	if err != nil {
+		g.logger.Error(ctx, "Failed to get egress list from GitHub, falling back to local list: %v", err)
+
+		egress, err := g.GetLocalEgressList()
+		if err != nil {
+			return "", "", err
+		}
+
+		return g.EgressListToString(egress, g.Variables)
+	}
+
+	egress, err := egressResponse.GetContent()
+	if err != nil {
+		return "", "", err
+	}
+
+	g.logger.Info(ctx, "Using egress URL list from %s at SHA %s", egressResponse.GetURL(), egressResponse.GetSHA())
+
+	return g.EgressListToString(egress, g.Variables)
+}
+
+func (g *Generator) GetLocalEgressList() (string, error) {
+	switch g.PlatformType {
 	case cloud.GCPClassic:
 		return templateGCPClassic, nil
 	case cloud.AWSHCP:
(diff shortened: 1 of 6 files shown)


pkg/data/egress_lists/egress_lists.go
```diff
@@ -1,21 +1,14 @@
 package egress_lists
 
-// TRANSITIONAL IMPLEMENTATION (UNSTABLE API)
```
joshbranham, line 3: This ticket was closed a while ago, this comment is no longer valid.

pkg/verifier/aws/entry_point.go
```diff
@@ -101,55 +100,33 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
...
-
-			a.Logger.Info(vei.Ctx, "Using egress URL list from %s at SHA %s", githubEgressList.GetURL(), githubEgressList.GetSHA())
-		}
-	}
+	// Generate both egress lists for the given PlatformType. Note: the result of this is ignored by the Legacy probe.
+	generatorVariables := map[string]string{"AWS_REGION": a.AwsClient.Region}
+	generator := egress_lists.NewGenerator(vei.PlatformType, generatorVariables, a.Logger)
 
-	egressListStr, tlsDisabledEgressListStr, err = egress_lists.EgressListToString(egressListYaml, map[string]string{"AWS_REGION": a.AwsClient.Region})
+	egressListStr, tlsDisabledEgressListStr, err := generator.GenerateEgressLists(vei.Ctx, vei.EgressListYaml)
```
joshbranham, line 129: The extraction of the fetching login into `egress_lists` essentially allows us to have this common call flow across AWS + GCP, and reads more cleanly in the scope of this already large function.

pkg/data/egress_lists/egress_lists.go
```diff
@@ -46,18 +87,14 @@ func GetLocalEgressList(platformType cloud.Platform) (string, error) {
...
+		return "", fmt.Errorf("no egress list registered for platform '%s'", g.PlatformType)
 	}
 }
 
-func GetGithubEgressList(platformType cloud.Platform) (*github.RepositoryContent, error) {
-	ghClient := github.NewClient(nil)
+func (g *Generator) GetGithubEgressList(ctx context.Context) (*github.RepositoryContent, error) {
 	path := "/pkg/data/egress_lists/"
-	if !platformType.IsValid() {
-		fmt.Printf("Platform type %s is invalid", platformType)
```
joshbranham, line 57: As described in the PR body, I removed this check, which was only printing the invalid platform and then continuing execution.

pkg/data/egress_lists/egress_lists.go
```diff
@@ -31,12 +24,60 @@ var templateGCPClassic string
...
+type githubReposClient interface {
+	GetContents(ctx context.Context, owner, repo, path string, opts *github.RepositoryContentGetOptions) (fileContent *github.RepositoryContent, directoryContent []*github.RepositoryContent, resp *github.Response, err error)
+}
+
+type Generator struct {
+	PlatformType      cloud.Platform
+	Variables         map[string]string
+	logger            logging.Logger
+	githubReposClient githubReposClient
+}
```
abyrne55, line 36: Could we add a docstring here, and perhaps change the name to something less generic? `EgressListGenerator` perhaps?

pkg/data/egress_lists/egress_lists.go
```diff
@@ -31,12 +24,60 @@ var templateGCPClassic string
...
+
+func NewGenerator(platformType cloud.Platform, variables map[string]string, logger logging.Logger) *Generator {
+	return &Generator{
+		PlatformType:      platformType,
+		Variables:         variables,
+		logger:            logger,
+		githubReposClient: github.NewClient(nil).Repositories,
+	}
+}
```
abyrne55, line 45: I'm curious about utility of a factory method like this for an exported struct. If the intent is provide a framework for type validation, couldn't someone get around this by just instantiating a `Generator` directly?

pkg/data/egress_lists/egress_lists.go
joshbranham: I chose the name `Generator` so it lines up nicely with the package name. `egress_lists.Generator` is pretty clear in the code, versus `egress_lists.EgressListGenerator` is a bit verbose, but I am happy to change it if you feel strongly

pkg/data/egress_lists/egress_lists.go
joshbranham: The intent is to provide a common way to create a `Generator` without having to think about the implementation details needed to call GitHub via our newly added interface. If some creates a `Generator` directly outside of this package they will end up getting a panic when trying to get GitHub egress lists (which I could fix for extra safety with error handling). I am open to other suggestions, I just find factory functions easier to reason about what is needed for a struct to work, and provides an easy mechanism to add further input validation in the future.

pkg/data/egress_lists/egress_lists.go
abyrne55: Ahh right, I always forget that everything gets prefixed with the package name. Nope, that makes perfect sense. Just slap on some comments for the type and its fields and we're good

pkg/data/egress_lists/egress_lists.go
abyrne55: Thanks for the explanation! That makes sense to me

abyrne55:
Successfully tested locally. Good stuff @joshbranham 

/approve

pkg/data/egress_lists/egress_lists.go
dakotalongRH: Do we want to change this to short circuit the execution at this point if the platform type is invalid?  Worth noting there may be earlier checks before it gets this far that would short circuit already

joshbranham: We provide the valid platform types as global variables, and the method `ByName(name string) (string, error)`, so it feels unnecessary to also then later validate a `Platform` is valid since callers should be utilizing the variables or `ByName...` to construct the type. This is further evident by the `names` field of the struct being private so it wouldn't be possible to construct an invalid one anyways afaict.

abyrne55: > it wouldn't be possible to construct an invalid one anyways afaict
There are several points in the codebase where we return an empty (and therefore invalid) `Platform{}`

joshbranham: > > it wouldn't be possible to construct an invalid one anyways afaict
> 
> There are several points in the codebase where we return an empty (and therefore invalid) `Platform{}`

Based on my searching, this is only happening under error cases inside other functions and in tests. If an upstream caller is not handling that error, and then using the `Platform{}` then that is a bug that should be fixed. Was there another example you had of this (outside of tests and errors)?

dakotalongRH: Correct, there are currently no other cases currently where `Platform{}` is used in network-verifier directly but it is possible to return an invalid type (with error) through `platform.ByName()` or `platform.IsValid()`.  As long as upstream callers handle that error I'm fine conceding the point and don't think the check is necessary here.  As far as I can tell the `ValidateEgress` func handles that beforehand.

joshbranham: > Correct, there are currently no other cases currently where `Platform{}` is used in network-verifier directly but it is possible to return an invalid type (with error) through `platform.ByName()` or `platform.IsValid()`. As long as upstream callers handle that error I'm fine conceding the point and don't think the check is necessary here. As far as I can tell the `ValidateEgress` func handles that beforehand.

Totally, and yup we handle the error validation already [here](https://github.com/openshift/osd-network-verifier/blob/2b020def9f2d5c2f4068d5d2804c6496c76795ce/pkg/verifier/aws/entry_point.go#L37-L39) in `ValidateEgress` (for AWS as an example)

PR #285, opened 2024-11-04
ignore any passed caCert when noTls option is set to true.

luis-falcon:
BUG fix

Ignore caCert option passed if the noTls flag was set to true.

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -91,16 +91,22 @@ are set correctly before execution.
 			}
 
 			// Set Up Proxy
-			if config.CaCert != "" {
-				// Read in the cert file
-				cert, err := os.ReadFile(config.CaCert)
-				if err != nil {
-					fmt.Println(err)
-					os.Exit(1)
+			if config.noTls && config.CaCert != "" {
+				fmt.Println("--cacert value will be ignored since --no-tls was passed")
+			}
+
+			if !config.noTls {
+				if config.CaCert != "" {
+					// Read in the cert file
+					cert, err := os.ReadFile(config.CaCert)
+					if err != nil {
+						fmt.Println(err)
+						os.Exit(1)
+					}
+					// store string form of it
+					// this was agreed with sda that they'll be communicating it as a string.
+					config.CaCert = bytes.NewBuffer(cert).String()
 				}
-				// store string form of it
-				// this was agreed with sda that they'll be communicating it as a string.
-				config.CaCert = bytes.NewBuffer(cert).String()
 			}
 
 			p := proxy.ProxyConfig{
--- a/pkg/verifier/aws/entry_point.go
+++ b/pkg/verifier/aws/entry_point.go
@@ -156,6 +156,11 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 
 	// Generate the userData file
 	// As expand replaces all ${var} (using empty string for unknown ones), adding the env variables used in userdata.yaml
+	if vei.Proxy.NoTls {
+		a.Logger.Info(vei.Ctx, "NoTls enabled; ignoring any provided CA certs")
+		vei.Proxy.Cacert = ""
+	}
+
 	userDataVariables := map[string]string{
 		"AWS_REGION":       a.AwsClient.Region,
 		"VALIDATOR_IMAGE":  networkValidatorImage,
--- a/pkg/verifier/gcp/entry_point.go
+++ b/pkg/verifier/gcp/entry_point.go
@@ -91,6 +91,10 @@ func (g *GcpVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 
 	// Generate the userData file
 	// Expand replaces all ${var} (using empty string for unknown ones), adding the env variables used in startup-script.sh
+	if vei.Proxy.NoTls {
+		g.Logger.Info(vei.Ctx, "NoTls enabled; ignoring any provided CA certs")
+		vei.Proxy.Cacert = ""
+	}
 	userDataVariables := map[string]string{
 		"TIMEOUT":          vei.Timeout.String(),
 		"HTTP_PROXY":       vei.Proxy.HttpProxy,


cmd/egress/cmd.go
```diff
@@ -144,6 +144,11 @@ are set correctly before execution.
 
 				awsVerifier.Logger.Warn(context.TODO(), "Using region: %s", config.region)
 
+				if vei.Proxy.NoTls && vei.Proxy.Cacert != "" {
```
joshbranham, line 147: I think it would make more sense for this to be done before line 94, so it applies to both AWS and GCP verifier blocks. For example, we load and parse the CA on line from the CLI flag, so instead we can just check `config.noTls` instead and skip the parsing of the CA altogether.

cmd/egress/cmd.go
```diff
@@ -144,6 +144,11 @@ are set correctly before execution.
...
+					awsVerifier.Logger.Warn(context.TODO(), "CaCert: %s, will not be used since --NoTls flag was used", vei.Proxy.Cacert)
```
joshbranham, line 148: A more succinct log like this can also be pulled up with my other suggestion

```
awsVerifier.Logger.Warn(context.TODO(), "--cacert value ignored since --no-tls was passed")
```

joshbranham (changes requested):

cmd/egress/cmd.go
luis-falcon: Good suggestion. Done

cmd/egress/cmd.go
luis-falcon: done

luis-falcon (commented):
Done

cmd/egress/cmd.go
```diff
@@ -91,16 +91,19 @@ are set correctly before execution.
...
 			// Set Up Proxy
-			if config.CaCert != "" {
-				// Read in the cert file
-				cert, err := os.ReadFile(config.CaCert)
-				if err != nil {
-					fmt.Println(err)
-					os.Exit(1)
+			if !config.noTls {
+
+				if config.CaCert != "" {
```
joshbranham, line 96: This is totally fine as is, but I think this could read more easily as something like this, with the warning being printed in this block as well versus later, what do you think?
```
if config.CaCert != "" && !config.noTls {
   // parse the cert logic
} else {
   fmt.Println("--cacert value will be ignored since --no-tls was passed")
}
```

cmd/egress/cmd.go
```diff
@@ -123,6 +126,10 @@ are set correctly before execution.
 				Proxy:        p,
 			}
 
+			if config.noTls && config.CaCert != "" {
+				fmt.Println("--cacert value will be ignored since --no-tls was passed")
+			}
```
joshbranham, line 131: Can we move this check to be above line 94 (where we first check `!config.noTls`)?

cmd/egress/cmd.go
```diff
@@ -91,16 +91,19 @@ are set correctly before execution.
...
 
 			// Set Up Proxy
-			if config.CaCert != "" {
-				// Read in the cert file
-				cert, err := os.ReadFile(config.CaCert)
-				if err != nil {
-					fmt.Println(err)
-					os.Exit(1)
+			if !config.noTls {
+
```
joshbranham, line 95:
```suggestion
```

joshbranham (changes requested):
A couple small suggestions to wrap just to make it a bit easier to read when `config.CaCert` input value is used.

joshbranham (commented):
Based on the ticket, this looks good at solving both callers using this module in other projects as well as via our CLI. Let's get @abyrne55 to take a peek to make sure we are not missing anything from the perspective of the ticket since he wrote it.

As discussed offline, ideally we would be able to factor things out a bit to make it impossible to construct an invalid `Proxy` in the first place with these mutually exclusive fields, but based on things are at this time this is a good solution.

/lgtm

pkg/verifier/aws/entry_point.go
```diff
@@ -156,6 +156,11 @@ func (a *AwsVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 
 	// Generate the userData file
 	// As expand replaces all ${var} (using empty string for unknown ones), adding the env variables used in userdata.yaml
+	if vei.Proxy.NoTls && vei.Proxy.Cacert != "" {
+		a.Logger.Info(vei.Ctx, " Cacert value will be ignored since NoTls is enabled.")
+		vei.Proxy.Cacert = ""
+	}
```
abyrne55, line 162:
```suggestion
	if vei.Proxy.NoTls {
		a.Logger.Info(vei.Ctx, "NoTls enabled; ignoring any provided CA certs")
		vei.Proxy.Cacert = ""
	}
```
If NoTls is specified, we don't care whether `vei.Proxy.Cacert` is blank or not; we're gonna ignore it. So why not simplify further?

pkg/verifier/gcp/entry_point.go
```diff
@@ -91,6 +91,10 @@ func (g *GcpVerifier) ValidateEgress(vei verifier.ValidateEgressInput) *output.O
 
 	// Generate the userData file
 	// Expand replaces all ${var} (using empty string for unknown ones), adding the env variables used in startup-script.sh
+	if vei.Proxy.NoTls && vei.Proxy.Cacert != "" {
+		g.Logger.Info(vei.Ctx, " Cacert value will be ignored since NoTls is enabled.")
+		vei.Proxy.Cacert = ""
+	}
```
abyrne55, line 97:
```suggestion
	if vei.Proxy.NoTls {
		g.Logger.Info(vei.Ctx, "NoTls enabled; ignoring any provided CA certs")
		vei.Proxy.Cacert = ""
	}
```
See other comment

abyrne55 (changes requested):
Nice work @luis-falcon! Just suggested a couple simplifications

pkg/verifier/gcp/entry_point.go
luis-falcon: true. Done

pkg/verifier/aws/entry_point.go
luis-falcon: Done

joshbranham (commented):
/lgtm

PR #287, opened 2024-11-05
OSD-26102 added architechturebyname check

luis-falcon:
added ArchitechtureByName() function check after normalization in order to account for an architecture with less than 3 names in the future. Also added better --notls and --cacert flags as mutually exclusive

- Bug
- Enhancement

### DIFF
--- a/cmd/egress/cmd.go
+++ b/cmd/egress/cmd.go
@@ -90,23 +90,16 @@ are set correctly before execution.
 				config.region = getDefaultRegion(platformType)
 			}
 
-			// Set Up Proxy
-			if config.noTls && config.CaCert != "" {
-				fmt.Println("--cacert value will be ignored since --no-tls was passed")
-			}
-
-			if !config.noTls {
-				if config.CaCert != "" {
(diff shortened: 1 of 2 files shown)


cmd/egress/cmd.go
```diff
@@ -275,6 +273,7 @@ are set correctly before execution.
 		validateEgressCmd.PrintErr(err)
 	}
 
+	validateEgressCmd.MarkFlagsMutuallyExclusive("cacert", "no-tls")
```
joshbranham, line 276: Now that we have this, the check and log on line 94-96 should be removed, as it is not possible to hit that codepath.

joshbranham (commented):
Just one bit to cleanup, and one nit: can you add the Jira ticket to the PR title like `OSD-1234: <message>`

cmd/egress/cmd.go
luis-falcon: i already removed that check

joshbranham (commented):
/lgtm

PR #288, opened 2024-11-13
osd-21895 network verifier highlights kms permission issues

luis-falcon:
Added code to allow network verifier to highlight issues when instance changes state to termination due to issues with KMS

- Enhancement/Feature

Solves https://issues.redhat.com/browse/OSD-21895

### DIFF
--- a/pkg/verifier/aws/aws_verifier.go
+++ b/pkg/verifier/aws/aws_verifier.go
@@ -71,7 +71,7 @@ const (
 	networkValidatorRepo  = "quay.io/app-sre/osd-network-verifier"
 	userdataEndVerifier   = "USERDATA END"
 	prepulledImageMessage = "Warning: could not pull the specified docker image, will try to use the prepulled one"
-	invalildKMSCode       = "Client.InvalidKMSKey.InvalidState"
+	invalidKMSCode       = "Client.InvalidKMSKey.InvalidState"
 )
 
 // AwsVerifier holds an aws client and knows how to fulfill the VerifierService which contains all functions needed for verifier
@@ -353,7 +353,7 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
 		}
 
 		waiterErr := fmt.Errorf("%s: terminated %s after timing out waiting for instance to be running", err, instanceID)
-		if stateCode == invalildKMSCode {
+		if stateCode == invalidKMSCode {
 			waiterErr = handledErrors.NewKmsError("encountered issue accessing KMS key when launching instance.")
 		}
 
--- a/pkg/verifier/aws/aws_verifier.go
+++ b/pkg/verifier/aws/aws_verifier.go
@@ -16,7 +16,6 @@ import (
 	ec2Types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/go-playground/validator"
 	ocmlog "github.com/openshift-online/ocm-sdk-go/logging"
-
 	"github.com/openshift/osd-network-verifier/pkg/clients/aws"
 	"github.com/openshift/osd-network-verifier/pkg/data/cloud"
 	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
@@ -72,6 +71,7 @@ const (
 	networkValidatorRepo  = "quay.io/app-sre/osd-network-verifier"
 	userdataEndVerifier   = "USERDATA END"
 	prepulledImageMessage = "Warning: could not pull the specified docker image, will try to use the prepulled one"
+	invalildKMSCode       = "Client.InvalidKMSKey.InvalidState"
 )
 
 // AwsVerifier holds an aws client and knows how to fulfill the VerifierService which contains all functions needed for verifier
@@ -340,10 +340,28 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
 	// Wait up to 5 minutes for the instance to be running
 	waiter := ec2.NewInstanceRunningWaiter(a.AwsClient)
 	if err := waiter.Wait(input.ctx, &ec2.DescribeInstancesInput{InstanceIds: []string{instanceID}}, 2*time.Minute); err != nil {
+		resp, err := a.AwsClient.DescribeInstances(context.TODO(), &ec2.DescribeInstancesInput{
+			InstanceIds: []string{instanceID},
+		})
+		if err != nil {
+			fmt.Println("Warning: Waiter Describe instances failure.")
+		}
+
+		var stateCode string
+		if resp.Reservations[0].Instances[0].StateReason.Code != nil {
+			stateCode = *resp.Reservations[0].Instances[0].StateReason.Code
+		}
+
+		waiterErr := fmt.Errorf("%s: terminated %s after timing out waiting for instance to be running", err, instanceID)
+		if stateCode == invalildKMSCode {
+			waiterErr = handledErrors.NewKmsError("encountered issue accessing KMS key when launching instance.")
+		}
+
 		if err := a.AwsClient.TerminateEC2Instance(input.ctx, instanceID); err != nil {
 			return instanceID, handledErrors.NewGenericError(err)
 		}
-		return "", fmt.Errorf("%s: terminated %s after timing out waiting for instance to be running", err, instanceID)
+
+		return "", waiterErr
 	}
 
 	return instanceID, nil
--- a/pkg/verifier/aws/aws_verifier.go
+++ b/pkg/verifier/aws/aws_verifier.go
@@ -71,7 +71,7 @@ const (
 	networkValidatorRepo  = "quay.io/app-sre/osd-network-verifier"
 	userdataEndVerifier   = "USERDATA END"
 	prepulledImageMessage = "Warning: could not pull the specified docker image, will try to use the prepulled one"
-	invalidKMSCode       = "Client.InvalidKMSKey.InvalidState"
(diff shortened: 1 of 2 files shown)


pkg/verifier/aws/aws_verifier.go
```diff
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	awsTools "github.com/aws/aws-sdk-go-v2/aws"
+
```
joshbranham, line 15:
```suggestion
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -339,11 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
 
 	// Wait up to 5 minutes for the instance to be running
 	waiter := ec2.NewInstanceRunningWaiter(a.AwsClient)
+
```
joshbranham, line 343:
```suggestion
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -339,11 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
...
+		resp, err := a.AwsClient.DescribeInstances(context.TODO(), &ec2.DescribeInstancesInput{
+			InstanceIds: []string{instanceID},
+		})
+		if err != nil {
+			fmt.Println("Warning: Waiter Describe instances failure.")
+		}
+
+		stateCode := *resp.Reservations[0].Instances[0].StateReason.Code
+		stateMessage := *resp.Reservations[0].Instances[0].StateReason.Message
```
joshbranham, line 356: Each of these has the potential to cause a nil pointer. We should check they are not nil before getting the value, with something like this

```go
 var stateCode, stateMessage string
 if resp.Reservations[0].Instances[0].StateReason.Code != nil {
   stateCode = *resp.Reservations[0].Instances[0].StateReason.Code
 }
 if resp.Reservations[0].Instances[0].StateReason.Message != nil {
   stateMessage = *resp.Reservations[0].Instances[0].StateReason.Message
 }
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -339,11 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
...
+
+		if stateCode == "Client.InvalidKMSKey.InvalidState" {
```
joshbranham, line 358: We should throw this string in a constant

pkg/verifier/aws/aws_verifier.go
```diff
@@ -339,11 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
...
+			waiterErr = fmt.Errorf("Instance %s terminated: %s. ", instanceID, stateMessage)
```
joshbranham, line 359: Let's make this a bit more concise, something like
```suggestion
			waiterErr = fmt.Errorf("instance %s terminated because of: %s. ", instanceID, stateMessage)
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -339,11 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
 
 	// Wait up to 5 minutes for the instance to be running
 	waiter := ec2.NewInstanceRunningWaiter(a.AwsClient)
+
 	if err := waiter.Wait(input.ctx, &ec2.DescribeInstancesInput{InstanceIds: []string{instanceID}}, 2*time.Minute); err != nil {
+
+		waiterErr := fmt.Errorf("%s: terminated %s after timing out waiting for instance to be running", err, instanceID)
```
joshbranham, line 346: If we are going to conditionally set this value, lets move the definition to be closer to where it is used and set, so down on line 357

joshbranham (changes requested):
The conditional logic to handle the KMS issues looks good to me, but lets clean things up a bit and protect against nil pointers.

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

luis-falcon (commented):
All comments on the review resolved.

pkg/verifier/aws/aws_verifier.go
```diff
@@ -340,10 +340,33 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
...
+			stateCode = *resp.Reservations[0].Instances[0].StateReason.Code
+		}
+		if resp.Reservations[0].Instances[0].StateReason.Message != nil {
+			stateMessage = *resp.Reservations[0].Instances[0].StateReason.Message
+		}
+
+		waiterErr := fmt.Errorf("%s: terminated %s after timing out waiting for instance to be running", err, instanceID)
+
+		if stateCode == invalildKMSCode {
+			waiterErr = fmt.Errorf("Instance %s terminated because of: %s. ", instanceID, stateMessage)
```
joshbranham, line 362: After reviewing the OSD ticket, it calls out we need to reach out to OCM and others to inform them of this new error message. So we should be certain of the wording here, and check with how they handle current errors (do they string match? do they wanted a typed error?) cc @abyrne55

luis-falcon: we returning exactly the same as before. I don't think this changes how OCM handles the error. I believe the communication is after the merge is done to communicate with the OCM team that there is this new feature. @abyrne55  Input?

joshbranham: It's more that under this specific failure scenario (KMS key access issues) we return a more descriptive error message now than the generic failure message before. If they want to bubble that up nicely to the customer in the API and UI, then how we implement it matters.

joshbranham: As discussed in Slack, here is a sample of how we could handle this to make it easier for OCM to know it was a KMS failure and provide a more helpful error message to customers using the API or Console. https://github.com/joshbranham/osd-network-verifier/pull/1

pkg/verifier/aws/aws_verifier.go
```diff
@@ -340,10 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
 	// Wait up to 5 minutes for the instance to be running
 	waiter := ec2.NewInstanceRunningWaiter(a.AwsClient)
 	if err := waiter.Wait(input.ctx, &ec2.DescribeInstancesInput{InstanceIds: []string{instanceID}}, 2*time.Minute); err != nil {
+
```
joshbranham, line 343:
```suggestion
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -340,10 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
...
+			fmt.Println("Warning: Waiter Describe instances failure.")
+		}
+
+		var stateCode string
+		if resp.Reservations[0].Instances[0].StateReason.Code != nil {
+			stateCode = *resp.Reservations[0].Instances[0].StateReason.Code
+		}
+
+		waiterErr := fmt.Errorf("%s: terminated %s after timing out waiting for instance to be running", err, instanceID)
+
```
joshbranham, line 357:
```suggestion
```

pkg/verifier/aws/aws_verifier.go
```diff
@@ -340,10 +340,30 @@ func (a *AwsVerifier) createEC2Instance(input createEC2InstanceInput) (string, e
...
+		if stateCode == invalildKMSCode {
+			waiterErr = handledErrors.NewKmsError("Encountered issue accessing KMS key when launching instance.")
```
joshbranham, line 359:
```suggestion
			waiterErr = handledErrors.NewKmsError("encountered issue accessing KMS key when launching instance.")
```

joshbranham (commented):
Just a couple `nit`s in terms of formatting/newlines, otherwise this approach LGTM

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

pkg/verifier/aws/aws_verifier.go
luis-falcon: Done

abyrne55 (approved):
Tested locally successfully. Nice work @luis-falcon, and thank you for your input @joshbranham

/lgtm
/approve
/unhold

PR #289, opened 2024-11-18
OSD-26236 leaver process member removal

luis-falcon:
Removing michael shen from OWNER FILE as part of the leaver process


joshbranham (approved):
/lgtm
/approve

PR #291, opened 2024-11-21
Remove servicequota URL from aws-classic.yaml

abyrne55:
This PR removed the servicequota.$REGION.amazonaws.com URL from the OSD/ROSA Classic egress URL list, as that URL has been deemed unnecessary and is about to be removed from the docs (see openshift/openshift-docs#77271)


joshbranham (approved):
/lgtm

PR #292, opened 2025-01-06
OSD-25041: Ensure we do the SecurityGroup swap for instance cleanup on KMS failures

joshbranham:

In https://issues.redhat.com/browse/OSD-25041 it was brought up that in CI, there are some non-zero number of failures where the KMS specific test fails and leaks the `osd-network-verifier` SecurityGroup. After reviewing things a bit more, we can see that when a KMS failure occurs, we exit early and terminate the instance. However, further up the call chain we would pre-detach the `osd-network-verifier` SG and attach the VPC default _before_ terminating to prevent any race condition.

This extracts that behavior a bit better and ensures we use it in both places.

## Checklist

## Reviewer's Checklist


abyrne55 (approved):
Looks good and passed my local tests. Great work @joshbranham!

/lgtm
/approve

PR #295, opened 2025-01-20
Properly set proxy env vars in systemd curl probe

abyrne55:
This PR fixes OSD-27707, which identified a bug in how we were passing the `http_proxy` and related environmental variables to the systemd-based curl probe. Using the `Environment` systemd directive ensures the variables are properly set


joshbranham (commented):
/hold
/lgtm

dakotalongRH:
/lgtm

Attaching logs from a NV run using the proxy
[nv_logs.txt](https://github.com/user-attachments/files/18481862/nv_logs.txt)

PR #310, opened 2025-05-13
Rewrite releasing doc

abyrne55:
This PR rewrites the releasing documentation (RELEASE.md) to better reflect current practices. This doc should also make it easier for maintainers to identify parts of the release process that could be better automated. In the meantime, this doc introduces some partial automation in the form of BASH snippets. Notably, this doc also formalizes the requirement that we enable deregistration protection on each AMI that's referenced in a released version.


joshbranham (commented):
This is a nice improvement, thanks for doing this. I think a roadmap item to wrap all this up in some wrapper script in this repo `make do-the-thing` that calls `go run script/thing.go` will be even nicer, then all this markdown logic becomes codified

/lgtm

abyrne55:
Thanks for the review @joshbranham. Just as a heads up, the default merge method on this repo is squash (per openshift/release#48901), so no need to set the tide label

PR #313, opened 2025-06-02
SREP-597: Export curl command building to another package

luis-falcon:
- Refactoring
Centralize the logic/template that we use to generate curl commands in the network verifier such that there are not multiple pieces of the codebase that need to be updated each time we need to make a change to the curl command.

Changed the instances in which we use the flag "-k" for its more readable equivalent "--insecure"

Exports the building of the Curl command used in the network verifier curl probe. into it's own package.

### DIFF
--- a/pkg/data/curlgen/curlgen.go
+++ b/pkg/data/curlgen/curlgen.go
@@ -1,0 +1,57 @@
+package curlgen
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// Options struct contains flag options that will be used to build
+// the curl command. Every Option directly maps to a flag that can
+// be used to configure your curl command.
+type Options struct {
+	CaPath          string
+	ProxyCaPath     string
+	Retry           int
+	MaxTime         string
+	NoTls           string
+	Urls            string
+	TlsDisabledUrls string
+}
+
+// GenerateString function will be used to transform the Configurations (options)
+// used to built the Options struct and build a full Curl command and return it as a string
+func GenerateString(cfg *Options, outputLinePrefix string) (string, error) {
+	command := fmt.Sprintf(`curl --capath %s --proxy-capath %s --retry %v --retry-connrefused -t B -Z -s -I -m %s -w "%%{stderr}%s%%{json}\n"`,
+		cfg.CaPath,
+		cfg.ProxyCaPath,
+		cfg.Retry,
+		cfg.MaxTime,
+		outputLinePrefix,
+	)
+
+	if cfg.NoTLS() {
+		command += " --insecure"
+		// In addition to adding the curl flag, we can merge the list of "tlsDisabled" URLs
+		// with the list of "normal" URLs now (since all URLs will be "tlsDisabled")
+		cfg.Urls += " " + cfg.TlsDisabledUrls
+		cfg.TlsDisabledUrls = ""
+	}
+
+	command += " " + cfg.Urls + " --proto =http,https,telnet"
+
+	if cfg.TlsDisabledUrls != "" {
+		command += fmt.Sprintf(
+			` --next --insecure --retry %v --retry-connrefused -s -I -m %s -w "%%{stderr}@NV@%%{json}\n" %s --proto =https`,
+			cfg.Retry,
+			cfg.MaxTime,
+			cfg.TlsDisabledUrls,
+		)
+	}
+
+	return command, nil
+
+}
+func (o *Options) NoTLS() bool {
+	noTLS, _ := strconv.ParseBool(o.NoTls)
+	return noTLS
+}
--- a/pkg/data/curlgen/curlgen_test.go
+++ b/pkg/data/curlgen/curlgen_test.go
@@ -1,0 +1,70 @@
+package curlgen
+
+import (
+	"testing"
+)
+
+func TestGenerateString(t *testing.T) {
+
+	tests := []struct {
+		name    string
+		args    *Options
+		want    string
+		wantErr bool
+	}{
+		{
+			name: "Happy Path",
+			args: &Options{
+				CaPath:          "/some/config/path/",
+				ProxyCaPath:     "/some/config/path/",
+				Retry:           3,
+				MaxTime:         "4",
+				NoTls:           "false",
+				Urls:            "http://example.com:80 https://example.org:443",
+				TlsDisabledUrls: "http://example2.com:80 https://example2.org:443",
+			},
+			want:    "curl --capath /some/config/path/ --proxy-capath /some/config/path/ --retry 3 --retry-connrefused -t B -Z -s -I -m 4 -w \"%{stderr}@NV@%{json}\\n\" http://example.com:80 https://example.org:443 --proto =http,https,telnet --next --insecure --retry 3 --retry-connrefused -s -I -m 4 -w \"%{stderr}@NV@%{json}\\n\" http://example2.com:80 https://example2.org:443 --proto =https",
+			wantErr: false,
+		},
+		{
+			name: "NoTls true with tlsDisabledUrls",
+			args: &Options{
+				CaPath:          "/some/config/path/",
+				ProxyCaPath:     "/some/config/path/",
+				Retry:           3,
+				MaxTime:         "4",
+				NoTls:           "true",
+				Urls:            "http://example.com:80 https://example.org:443",
+				TlsDisabledUrls: "http://example2.com:80 https://example2.org:443",
+			},
+			want:    "curl --capath /some/config/path/ --proxy-capath /some/config/path/ --retry 3 --retry-connrefused -t B -Z -s -I -m 4 -w \"%{stderr}@NV@%{json}\\n\" --insecure http://example.com:80 https://example.org:443 http://example2.com:80 https://example2.org:443 --proto =http,https,telnet",
+			wantErr: false,
+		},
+		{
+			name: "NoTls False with No TlsDisabledURLs",
+			args: &Options{
+				CaPath:          "/some/config/path/",
+				ProxyCaPath:     "/some/config/path/",
+				Retry:           3,
+				MaxTime:         "4",
+				NoTls:           "false",
+				Urls:            "http://example.com:80 https://example.org:443",
+				TlsDisabledUrls: "",
+			},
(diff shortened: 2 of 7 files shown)


pkg/probes/curl/curl_json_test.go
```diff
@@ -78,7 +78,7 @@ func TestCurlJSONProbe_GetExpandedUserData(t *testing.T) {
 				"URLS":    "http://example.com:80 https://example.org:443",
 				"NOTLS":   "True",
 			},
-			wantRegex: `#cloud-config[\s\S]* -k [\s\S]*http:\/\/example.com:80 https:\/\/example.org:443`,
+			wantRegex: `#cloud-config[\s\S]* --insecure [\s\S]*http:\/\/example.com:80 https:\/\/example.org:443`,
```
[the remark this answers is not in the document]
luis-falcon: Yes. -k and --insecure are equivalent.

pkg/data/curlgen/curlgen.go
```diff
@@ -0,0 +1,56 @@
+package curlgen
+
+import (
+	"fmt"
+	"strconv"
+)
+
+type Options struct {
```
joshbranham, line 8: Can you add docs for this public struct?

pkg/data/curlgen/curlgen.go
```diff
@@ -0,0 +1,56 @@
...
+	ProxyCaPath     string
+	Retry           int
+	MaxTime         string
+	WriteOut        string
+	NoTls           string
+	Urls            string
+	TlsDisabledUrls string
+}
+
+func GenerateString(cfg Options) (string, error) {
```
joshbranham, line 19: Can you document this function?

pkg/data/curlgen/curlgen.go
```diff
@@ -0,0 +1,56 @@
+package curlgen
```
joshbranham, line 1: Needs tests

pkg/probes/curl/curl_json.go
```diff
@@ -151,33 +150,20 @@ func (clp Probe) GetExpandedUserData(userDataVariables map[string]string) (strin
...
-			// with the list of "normal" URLs now (since all URLs will be "tlsDisabled")
-			userDataVariables["URLS"] += " " + userDataVariables["TLSDISABLED_URLS"]
-			userDataVariables["TLSDISABLED_URLS"] = ""
-		}
+	curlCommand := curlgen.Options{
+		CaPath:          "/etc/pki/tls/certs/",
+		ProxyCaPath:     "/etc/pki/tls/certs/",
+		Retry:           3,
+		MaxTime:         userDataVariables["TIMEOUT"],
+		WriteOut:        fmt.Sprintf("%%{stderr}%s%%{json}\\n", outputLinePrefix),
```
joshbranham, line 158: What's the motivation for removing `LINE_PREFIX` from the userdata and doing this here?

pkg/probes/curl/curl_json_test.go
```diff
@@ -44,41 +44,45 @@ func TestCurlJSONProbe_GetExpandedUserData(t *testing.T) {
 		{
 			name: "happy path",
 			userDataVariables: map[string]string{
-				"TIMEOUT": "1",
-				"DELAY":   "2",
-				"URLS":    "http://example.com:80 https://example.org:443",
+				"TIMEOUT":      "1",
+				"DELAY":        "2",
+				"URLS":         "http://example.com:80 https://example.org:443",
+				"CURL_COMMAND": "curl",
```
joshbranham, line 50: This doesn't actually fix the code, just the tests right? https://github.com/joshbranham/osd-network-verifier/blob/c1120adf883b0593eadc59413878e0ca61c04e21/pkg/probes/curl/curl_json.go#L105 runs and expects `presetUserVariables` to have `CURL_COMMAND` because `userdata-template.yaml` states `CURL_COMMAND` as required. You need to either modify the validation logic, or define the curl command earlier to satisfy the validation.

joshbranham (changes requested):

pkg/probes/curl/curl_json.go
joshbranham: But we don't want the `"@NV@"` to be configurable, right? That is not about stderr or stdout, but the characters we use to capture when the output should be parsed.

pkg/probes/curl/curl_json.go
```diff
@@ -100,9 +99,26 @@ func (clp Probe) GetExpandedUserData(userDataVariables map[string]string) (strin
 	// Extract required variables specified in template (if any)
 	directivelessUserDataTemplate, requiredVariables := helpers.ExtractRequiredVariablesDirective(userDataTemplate)
 
+	curlCommand := curlgen.Options{
+		CaPath:          "/etc/pki/tls/certs/",
+		ProxyCaPath:     "/etc/pki/tls/certs/",
+		Retry:           3,
+		MaxTime:         userDataVariables["TIMEOUT"],
```
joshbranham, line 106: You can't just move this up, the data for `TIMEOUT` for example is normalized via a function call on line 112/128.

luis-falcon: Done with reorg logic. Normalize first then command generation then validation of required vars.

pkg/probes/curl/userdata-template.yaml
```diff
@@ -1,12 +1,12 @@
...
-# network-verifier-required-variables=TIMEOUT,DELAY,URLS
+# network-verifier-required-variables=CURL_COMMAND,DELAY
 ${CACERT_RENDERED}
 runcmd:
   - systemctl mask --now serial-getty@ttyS0.service
   - dmesg -D
   - echo "${USERDATA_BEGIN}" >/dev/ttyS0
   - export http_proxy=${HTTP_PROXY} https_proxy=${HTTPS_PROXY} no_proxy="${NO_PROXY}"
-  - curl --capath /etc/pki/tls/certs/ --proxy-capath /etc/pki/tls/certs/ --retry 3 --retry-connrefused -t B -Z -s -I -m ${TIMEOUT} -w "%{stderr}${LINE_PREFIX}%{json}\n" ${CURLOPT} ${URLS} --proto =http,https,telnet ${TLSDISABLED_URLS_RENDERED} 2>/dev/ttyS0
+  - ${CURL_COMMAND} 2>/dev/ttyS0
```
joshbranham, line 9: Should this be quoted to prevent shell expansion? Unsure if cloud-init does that for us or not, but worth confirming/testing.

joshbranham (changes requested):

pkg/data/curlgen/curlgen.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Options struct contains flag options that will be used to build
+// the curl command. Every Option directly maps to a flag that can
+// be used to configure your curl command.
+type Options struct {
+	CaPath          string
+	ProxyCaPath     string
+	Retry           int
+	MaxTime         string
+	NoTls           string
```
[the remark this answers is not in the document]
luis-falcon: It is already coming as a string from the userdataVariables["NOTLS"] which is coming from farther back. If I do this either I change the code from the start to be boolean from the beginning. or then i have to convert 2 times instead of one in the function.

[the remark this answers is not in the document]
joshbranham: Then a clever solution here to make this easier is to bury the bool conversion on the `Options` struct so you can call that easily. Something like the following, but not required.

```golang
func (o *Options) NoTLS() (bool) {
  // do the conversion/comparison here and return a bool
  noTLS, _ := strconv.ParseBool(o.NoTls)
  return noTLS
}
```

pkg/probes/curl/userdata-template.yaml
luis-falcon: If I do this. All the curl_json_tests.go unit tests fail.

pkg/data/curlgen/curlgen_test.go
```diff
@@ -0,0 +1,78 @@
...
+			name: "happy Path",
+			args: args{
+				cfg: Options{
+					CaPath:          "/some/config/path/",
+					ProxyCaPath:     "/some/config/path/",
+					Retry:           3,
+					MaxTime:         "4",
+					NoTls:           "true",
+					Urls:            "http://example.com:80 https://example.org:443",
+					TlsDisabledUrls: "http://example2.com:80 https://example2.org:443",
```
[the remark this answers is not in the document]
luis-falcon: Done

pkg/probes/curl/userdata-template.yaml
geowa4: Should `${CURL_COMMAND}` be quoted? Don't we _want_ shell expansion here? Line 7 quotes the variable but that's because it's an argument to echo. In Line 9's case, we replaced a command and its arguments with a variable. If we quote that variable, none of the arguments will expand. Here's my test in bash.

```
bash-5.2$ CURL_COMMAND='curl -sSLi'
bash-5.2$ "${CURL_COMMAND}" https://www.redhat.com
bash: curl -sSLi: command not found
```

Dropping the quotes runs the curl as expected.

pkg/data/curlgen/curlgen.go
```diff
@@ -0,0 +1,58 @@
...
+			cfg.Urls += " " + cfg.TlsDisabledUrls
+			cfg.TlsDisabledUrls = ""
+		}
+	}
+
+	command += " " + cfg.Urls + " --proto =http,https,telnet"
+
+	if cfg.TlsDisabledUrls != "" {
+		command += fmt.Sprintf(
+			` --next --insecure --retry %v --retry-connrefused -s -I -m %s -w "%%{stderr}@NV@%%{json}\n" %s --proto =https`,
```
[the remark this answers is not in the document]
geowa4: Looks like it is a constant: https://github.com/openshift/osd-network-verifier/pull/313/files#diff-84c05e9f2d765cd933d08d61c5e09ea3f2968ef25cc17c8bcbce6ce90988e5d8R40

pkg/data/curlgen/curlgen_test.go
```diff
@@ -0,0 +1,86 @@
+package curlgen
+
+import (
+	"testing"
+)
+
+const (
+	outputLine = "@NV@"
```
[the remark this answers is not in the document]
geowa4: https://github.com/openshift/osd-network-verifier/pull/313/files#diff-84c05e9f2d765cd933d08d61c5e09ea3f2968ef25cc17c8bcbce6ce90988e5d8R40

joshbranham (approved):
Ran the integration test locally and as best I can tell the existing verifier behavior is working after this logic is extracted.

/lgtm
/approve

PR #315, opened 2025-07-02
Added KubeAPI "cloud client" necessary for pod mode

abyrne55:
This PR implements SREP-674, which calls for a new "KubeAPI cloud client" to be implemented similarly to the existing GCP/AWS cloud clients so as to enable pod mode. The "meat" of this effort lies within pkg/clients/kube/kube.go, which offers a simplified interface for launching Jobs on k8s/OpenShift clusters and fetching their pods' output logs. 

This interface will primarily be used by pkg/verifier/kube/kube_verifier.go (currently just a stub) in order to launch curl Jobs and parse their logs into the existing "egressURL error" format that's expected by our downstream consumers. SREP-000 (TODO: cut card) will implement this logic. For now, I've added a "Hello World" demo to the osd-network-verifier standalone CLI that can be triggered using `--pod-mode` (after logging into an OpenShift cluster). The demo simply creates a Job in a hardcoded namespace (currently `openshift-network-diagnostics`), waits for it to complete, fetches its logs, and then deletes it.

Since this PR doesn't touch any of the code files/paths used by our downstream consumers, I'm posting it before I've had the chance to much more than bare-minimum testing on an OSD-on-AWS cluster. Haven't yet tested on ROSA [HCP/Classic] or OSD-GCP, although I don't expect anything to happen differently given the cloud-agnostic nature of pod mode. But by all means, go to town :)

## Checklist

## Reviewer's Checklist

```sh
ocm backplane login my-staging-cluster
make test build
./osd-network-verifier egress --pod-mode
```

## Logs 
```
$ ./osd-network-verifier egress --pod-mode
START POD MODE HELLO WORLD EXAMPLE
Created job; waiting for completion...
Job completed; getting logs...
Hello world from osd-network-verifier pod mode!

Deleting job...
Job deleted
END POD MODE HELLO WORLD EXAMPLE
```
The "Hello world from..." line is what's printed by the Job


pkg/clients/kube/kube.go
```diff
@@ -0,0 +1,149 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Client represents a Kubernetes Client
+type Client struct {
+	clientset kubernetes.Interface
+	namespace string
```
joshbranham, line 20: I would rather callers pass the namespace into functions than have to create a new client for each namespace.

pkg/clients/kube/kube.go
```diff
@@ -0,0 +1,149 @@
...
+				return fmt.Errorf("job %s failed", jobName)
+			}
+		}
+	}
+}
+
+// CleanupJob deletes a job and waits for the KubeAPI to let us know that it's gone
+// TODO: this seems to take many seconds longer to return than the output of
+// `oc get jobs -w --output-watch-events` implies the deletion takes. We should
+// investigate why this is the case.
```
joshbranham, line 123: Your comment or claude?

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/openshift/osd-network-verifier
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.23.6
+toolchain go1.24.2
```
joshbranham, line 5: Until we update boilerplate to 1.24 you will need to stay on 1.23.

cmd/egress/cmd.go
```diff
@@ -121,6 +129,68 @@ are set correctly before execution.
...
+				kubeVerifier, err := kubeverifier.NewKubeVerifier(clientset, config.debug)
+				if err != nil {
+					fmt.Printf("could not build kubeVerifier: %v\n", err)
+					os.Exit(1)
+				}
+
+				// HELLO WORLD EXAMPLE
+				fmt.Println("START POD MODE HELLO WORLD EXAMPLE")
+				kubeVerifier.KubeClient.SetNamespace("openshift-network-diagnostics")
+				kubeVerifier.KubeClient.CreateJob(context.TODO(), &batchv1.Job{
```
joshbranham, line 158: So this is where we would inject a `podmode` probe that generates the job then hands it in yeah?

joshbranham (commented):
Couple comments but looks pretty good!

pkg/clients/kube/kube.go
```diff
@@ -0,0 +1,149 @@
...
+	logs, err := c.clientset.CoreV1().Pods(c.namespace).GetLogs(pod.Name, &corev1.PodLogOptions{}).Do(ctx).Raw()
+	if err != nil {
+		return "", fmt.Errorf("failed to get logs for pod %s: %w", pod.Name, err)
+	}
+
+	return string(logs), nil
+}
+
+// WaitForJobCompletion waits for a job to complete (succeed or fail)
+func (c *Client) WaitForJobCompletion(ctx context.Context, jobName string, timeout time.Duration) error {
```
geowa4, line 89: Do we need to pass in both a context and a timeout or can the context that's passed in handle it? We don't have a way to cancel the WatchJob function since the passed in context is just `TODO`.

joshbranham: I have usually seen the following:
```
context.WithTimeout(context.Background(), time.Second*3)
```

joshbranham: Oh I see you are doing this on line 96 :)

abyrne55: We discussed this in handover last week, and the outcome was that we'll just enforce a timeout via a context that's configured by the caller instead of having a dedicated timeout parameter

PR #331, opened 2025-10-21
update explicit proxy example for HCP compatability 

eth1030:
Update proxy machine route table to allow all traffic in proxied subnet through proxy machine instead of just traffic from the proxy machine, scale up proxy machine instance so there are enough networking resources for HCP data plane installation, and add EC2 ELB S3 endpoints as required by ROSA installation in cluster-wide proxy.

## Checklist

## Reviewer's Checklist

Follow the explicit proxy example README

### DIFF
--- a/examples/aws/terraform/vpc-proxied-explicit/main.tf
+++ b/examples/aws/terraform/vpc-proxied-explicit/main.tf
@@ -167,10 +167,6 @@ resource "aws_route_table" "proxied" {
   vpc_id = aws_vpc.main.id
   tags   = { Name = "${var.name_prefix}-proxied-rtb" }
 
-  route {
-    cidr_block           = "0.0.0.0/0"
-    network_interface_id = aws_instance.proxy_machine.primary_network_interface_id
-  }
   dynamic "route" {
     for_each = toset(var.proxied_subnet_escape_routes)
     content {


examples/aws/terraform/vpc-proxied-explicit/main.tf
```diff
@@ -168,7 +168,7 @@ resource "aws_route_table" "proxied" {
   tags   = { Name = "${var.name_prefix}-proxied-rtb" }
 
   route {
-    cidr_block           = var.proxied_subnet_cidr_block
+    cidr_block           = "0.0.0.0/0"
```
joshbranham: I would want to confirm from Anthony the intent here. It looks like we are setting up a route for the whole subnet that the proxy is in, and then using `proxied_subnet_escape_routes` to let normal traffic go out to the internet. Were you using that variable? It's also possible I am misreading this.

abyrne55: Since this is the explicit proxy, you theoretically shouldn't need to force all traffic through the proxy machine via routing table, as each node should be voluntarily connecting to the proxy's IP. That being said, I'm not sure why this was originally set to route local/intra-subnet traffic into a machine that lives in a different subnet; I think that would break intra-subnet node-to-node comms? So you're on to something, @eth1030.

I think what the proxied subnet needs is:
* [a "local" route](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table#adopting-an-existing-local-route) (AWS has probably already been creating this for us, but it's not bad to make it explicit)
* a route allowing access to the proxied machine (could be a route for just its IP with /32, or it could be a route for the whole public subnet's CIDR range)
* escape routes, if configured

I don't think we need a route for 0.0.0.0/0

eth1030: Deleting the route so aws creates it automatically and all traffic is not forced through the proxy machine.

